PR #61, opened 2022-07-19
Add job proccessing

leepeuker:


leepeuker:
Todo: Disable sync buttons if trakt credentials are not set

Issue #219, opened 2023-01-09
Import Netflix CSV history files


JVT038:
Netflix supports exporting the viewing history. It can be exported [here](https://www.netflix.com/viewingactivity).

A screenshot of a formatted example (my viewing history):

![image](https://user-images.githubusercontent.com/47184046/211395759-36787531-94fa-45b2-908c-975421c3821c.png)

It would be useful to be able to import this into Movary.

leepeuker:
https://github.com/leepeuker/movary/releases/tag/0.42.0

JVT038:
@leepeuker So I'm looking for the next thing to work on, and I'm considering importing the Plex watch history.
Is this OK with you? I don't know which direction you want this project to go to and what the 'roadmap' is to the stable release.

My idea:
After implementing the Plex import, I'll look to import Jellyfin history and maybe importing the [Emby](https://emby.media/) history as well.
Hopefully the API of Jellyfin doesn't differ _too_ much from the Emby API and we can literally just use the same code for both APIs. 
When this importing stuff is done, I would like to add support for TV shows and ultimately add a dark mode to every page.

leepeuker:
The roadmap for V1 you can see here: https://github.com/leepeuker/movary/milestone/1
I will take time at the weekend to think about it again and make the tasks for v1 "final".

I am fine when you are working on the plex importer, that would have the highest prio of importers from my site.

I would say let us discuss the next step afterwards

JVT038:
I would definitely add importing Jellyfin history to the v1 milestone.
There are a lot of people in the self-hosting community (your target audience) who use Jellyfin. So this feature would be very nice for a lot of people.

I'll start working on the Plex importer and create a draft once I have something.

leepeuker:
I have created an issue for this topic https://github.com/leepeuker/movary/issues/272

For version 1.0 I want all currently known potential breaking changes and UX/UI changes (this includes things e.g. icon change) done. New features like jellyfin importer etc are not a requirement from my perspective but a nice to have for v1, I see no issue delivering this at a later point

PR #222, opened 2023-01-10
Moving the settings to a tab-structure

JVT038:
This isn't done yet, but I wanted to create a draft for some feedback.

Basically, I just moved the settings pages to the component folder, removed some code that would be included in the main settings base page (`settings.html.twig`) and included them in the main base.


JVT038:
I have two additional proposals regarding the settings page:

* Use [XHR](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest) (AJAX? It's already included in jQuery, soo...) instead of the current forms to submit data. The page currently reloads on each submission, which is kind of slow for me (and probably others). I would prefer XHR, as it would be faster.
* Use node modules. Using NPM or Yarn, we can work with frontend assets way more efficient, I think. It would decrease the size of this repository, and it would make asset compression using some kind of framework like React, Vite, Vue or Angular possible. 

Not related, but I would appreciate some kind of explanation for the application structure. Perhaps I'll make a documentation myself (if I have time), but I spent a while just figuring out what is happening, why it's happening and where it's happening. Some explanatory comments in the code would be appreciated. 

Also unrelated, but I wanted to compliment you on what seems to be an entire MVC structure without any framework. Yes, you've used some external libraries like FastRoute, Twig and a bit of symfony, but besides that, I think you've made the controllers yourself, the model, API, and views. Really impressive; I usually use Laravel for this.
But because you've made your own MVC structure, it's confusing to navigate to me. Especially considering I haven't seen any comments explaining anything.

leepeuker:
Hey, thanks for the PR and your time.

Please focus PRs on one scope of changes (here for the settings fronted) and remove everything not directly related (e.g. the docker and migrations changes), so that we can focus on one topic, makes discussions simpler. You can create separate PRs for the other things.

I preferred having the settings pages separate for a few reasons:
- frontend state management: much less complicated to handle the state of multiple small pages than one big page
- backend handling: currently for each page the backend only has to gather a little bit of data to render, all of that would have to be combined if there is only one settings page. Of course we could load the needed data later when the relevant tab is triggered, but this increases complexity in the frontend when on tab presses the tab data has to be prepared first.
- one bug can break more: when everything is on one page a error which would atm only break one settings subpage could break all settings (e.g. a error in the plex settings js file can only break the plex settings page atm). 
- smaller initial page load: Why load the data for all settings tabs when I am only interested in one of them? Sure if you click through all sub-setting-pages the basic layout has to be loaded every time, which adds a overhead, but you usually do not do that I think

I expect the content of the setting page/s to grow much more, so the more setting options there are the more the points I mentioned will gain in relevance I think.

Important Disclaimer: I am a backend guy and usually do not work with frontend stuff and have limited knowledge in this area, keep this in mind, my goal is most of the time to move as much complexity as possible in the backend because I know how to handle it there. For me a perfect frontend page contains little to no javascript and is as static as possible.
That is the reason why I prefer using form submits over XHR and why I am working with basic js here instead of some framework like React or similar, I do not have the time to get comfortable with this.

Someone with focus on frontend will probably disagree with most of what I said here :smile: 

> Not related, but I would appreciate some kind of explanation for the application structure. 

That is a good point, I will create an issue for this and look into it

> Also unrelated, but I wanted to compliment you on what seems to be an entire MVC structure without any framework [...] But because you've made your own MVC structure, it's confusing to navigate to me.

Thanks, I had bad experiences with symfony and laravel projects in the past (had to work around framework limitations, issues when upgrading dependencies and things like that), there are more than a couple of private projects I started which died because of frameworks for me. So I usually try to be bare bones about my projects, I only use what I really need to decrease the amount of external dependencies and to be as flexible as possible. I am not saying that that these frameworks are generally bad and especially in an open source setting like this it would probably make sense to use one, but I think I would have never reached this state of the project if I would have used one of these frameworks, I would probably have lost interest working on it long ago.

leepeuker:
Oh and I wanted to mention that the next big task I want to work on is to add clean REST API, so that e.g. the development of a more loosely coupled frontend is much easier possible than right now. I have cleaned up the current web api routes [here](https://github.com/leepeuker/movary/pull/223/files) in preparation for this.

JVT038:
> frontend state management: much less complicated to handle the state of multiple small pages than one big page

The different panes (for account, jellyfin, plex, etc.) are currently all contained in their own file in the `components` folder. So I don't think that it's more complicated to manage the pages with components. 

> one bug can break more: when everything is on one page a error which would atm only break one settings subpage could break all settings (e.g. a error in the plex settings js file can only break the plex settings page atm).

Then just make sure to test thorough and that nothing is broken.

> Important Disclaimer: I am a backend guy and usually do not work with frontend stuff and have limited knowledge in this area, keep this in mind, my goal is most of the time to move as much complexity as possible in the backend because I know how to handle it there. For me a perfect frontend page contains little to no javascript and is as static as possible.
That is the reason why I prefer using form submits over XHR and why I am working with basic js here instead of some framework like React or similar, I do not have the time to get comfortable with this.

I get that. I'm mostly a backend guy as well, because I honestly suck at designing stuff. However, I think learning more about frontend and how JS works can really increase your abilities, because you are able to do way more in the frontend and no more restricted to only backend. 

Movary is your project not mine, and you make the final decisions. 
If this PR gets closed, I'm going to work on #219, which shouldn't be too hard to implement.

JVT038:
Should I close this PR?

leepeuker:
Yes, I would say let us close this for now.

I have opened a new [issue](https://github.com/leepeuker/movary/issues/225) for the general topic frontend improvement

PR #228, opened 2023-01-12
Import netflix CSV file

JVT038:
Hi, this is the draft for importing activity history from Netflix.
I tried to stick to the Netflix stuff only this time.
So it's currently possible to go to the settings, navigate to the Netflix tab, select a csv file to import and view your imported csv file in a nice table.
In the backend, the file is first checked for any errors. This is why I added another HTTP response, [HTTP error 415](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/415) . Error 415 means that the user has submitted something in the wrong format, which is a nice response if the user uploads something that's not a csv file.
After validation, every single row in the csv file will be searched and checked whether it's a movie or a tv show. To check this, I had to use regex, because it's not directly stated. 
Currently, if an item is a tv show, it's ignored. If it's a movie, a search query will be sent to TMDb and the code will try to fetch the movie from TMDb. 
When all movies have been fetched from TMDb, the program returns a JSON string with all the data from TMDb and this data will be shown on the Netflix settings page, in a table.
The reason it's shown in a table instead of instantly imported into the database, is because searching TMDb isn't 100% accurate. Wrong movies may be picked up, and in this case the user has to manually edit this.

I'm planning to improve the frontend and to add an edit feature, so the user can manually edit any wrong records from TMDb. 
When the user is content with the imported items, they can click on an 'import button' which will import all the Netflix items into the datatabase, based on the TMDb data.
Let me know what you think of this.
I'm not really sure about the controller's location. Does Netflix need to become its own 'service'? Or is it fine as it is right now? I assumed that most functions in the controller won't be used outside the controller, and those functions can therefore be set as private.


leepeuker:
I had a quick look and could not find any big issue, the basic structure looks good. I would prefer a separate service class for the import logic, the controller classes should usually only extract data from requests, call the necessary services and create a response with the results.
I will look into the PR on more detail when I got more time in the next view days.

Oh but one thing, please move all changes to the docker setup and similar into a separate PR, so that we can discuss these if necessary outside of the Netflix import context.

JVT038:
I'll move the import logic to another service then.

For some reason, the docker mysql container doesn't work without the `user` stuff (due to permissions), which is why I added it. I'll try and see if I can fix it somehow.

templates/page/settings-netflix.html.twig
```diff
@@ -0,0 +1,69 @@
+{% extends 'base.html.twig' %}
+
+{% block title %}
+    Settings - Jellyfin
```
leepeuker, line 4:
```suggestion
    Settings - Netflix
```

src/ValueObject/Http/Response.php
```diff
@@ -24,6 +24,12 @@ public static function createBadRequest() : self
         return new self(StatusCode::createBadRequest());
     }
 
+    public static function createUnsupportedMediaType() : self
+    {
+        return new self(StatusCode::createUnsupportedMediaType());
+    }
+
+
```
leepeuker, line 32:
```suggestion
```

one blank line too much :P

.gitignore
```diff
@@ -1,7 +1,15 @@
...
 docker-compose.override.yml
+.phpunit.result.cache
+tmp/*
+.idea/
+vendor/
+.vscode
+.env
+/vendor
+/public/storage
+docker-compose.override.yml
```
leepeuker, line 15: there are a bunch of duplicated entries

src/Service/Netflix/ImportNetflixActivity.php
```diff
@@ -0,0 +1,112 @@
...
+            "(.+): .+ (\d{1,2}): (.*)",
+            // Check for TvShow : Season 1 - Part A: EpisodeTitle
+            "(.+): .+ (\d{1,2}) - .+: (.*)",
+            // Check for TvShow : TvShow : Miniseries : EpisodeTitle
+            "(.+): \w+: (.+)",
+            // Check for TvShow: SeasonName : EpisodeTitle
+            "(.+): (.+): (.+)",
+            // Check for TvShow: EpisodeTitle
+            "(.+): (.+)"
+        ];
```
JVT038, line 62: A lot of additional regex patterns still need to be added:

Check for TvShow: TvShow: SeasonName: EpisodeTitle: EpisodePart
Example: Avatar: The Last Airbender: Book 3: The Day of the Black Sun: Part 1

Check for TvShow: TvShow: SeasonName: EpisodeTitle
Example: Avatar: The Last Airbender: Book 3: The Painted Lady

Check for TvShow: Season 1: EpisodeTitle: Part 1
Example: Brooklyn Nine-Nine: Season 4: Coral Palms: Part 1

It's important to make a distinction between all the patterns, so that no pattern has overlap with another.

src/Service/Netflix/ImportNetflixActivity.php
```diff
@@ -0,0 +1,112 @@
...
+            // check for a pattern TvShow : Season 1: EpisodeTitle
+            // Example: Wednesday: Season 1: A Murder of Woes
+            "(.+): .+ (\d{1,2}): (.*)",
+            // Check for TvShow : Season 1 - Part A: EpisodeTitle
+            "(.+): .+ (\d{1,2}) - .+: (.*)",
+            // Check for TvShow : TvShow : Miniseries : EpisodeTitle
+            "(.+): \w+: (.+)",
+            // Check for TvShow: SeasonName : EpisodeTitle
+            "(.+): (.+): (.+)",
+            // Check for TvShow: EpisodeTitle
```
JVT038, line 60: A bit unreliable, because movie occasionally use the same format. (e.g "King Arthur: Legend of the Sword","17.01.21")
I'm currently not sure how to fix this, because there is no way to tell the difference between the two using regex.

JVT038:
So I'm currently thinking of the following workflow.

If an item has not been fetched from TMDB (because it wasn't found) or the wrong item from TMDB has been fetched, the user can change the item.
When the user clicks on the edit button, a modal opens with a search bar where the user can search for the correct item in the TMDB database.
The user searches for the item in TMDB, selects the right movie and the item is updated in the overview page.

Question: Where do I add the controller for search TMDB? Do I add it to the Netflix controller or should the request be forwarded to a TMDB controller?
I'll just add it to the Netflix controller for now. The request uri will be something like `/settings/netflix/search` with POST method.

leepeuker:
> I'll just add it to the Netflix controller for now

Fine by me for now, we can refactor this later if needed.

JVT038:
Question: I want to return the TMDB data in a JSON format, but I just saw that there's a class called `Movary\Util\Json` where there's a decode and encode function.
Should I use the built-in `json_encode()` or the `Json::encode()` to encode data into JSON?

leepeuker:
Use the json class, this guarantees the same decode/encode config in the codebase.

JVT038:
So I'm _almost_ done with this.
I have implemented the functionality to upload a file, filter out all the movies, search for those movies in TMDB, show the results in a nice table, and optionally edit the fetched TMDB items by searching for a different item.
I have also added a functionality to filter for items that have not been found by TMDB. Unfortunately, I didn't manage to add a page functionality to this, so I just made it so that all the not found items are shown in one page.
I'm expecting this to be done in a week or so.

public/js/settings-netflix.js
```diff
@@ -0,0 +1,494 @@
...
+
+		rating = ratingStarNumber;
+	}
+
+	return rating;
+}
+
+function updateRatingStars () {
+	setRatingStars(this.id.substring(20, 10));
+}
```
JVT038, line 488: This has almost been identically copied from `logMovie.js`, apart from one line.
I was wondering if we shouldn't just create a separate JS file and put the rating stuff in there.
Then we can include that JS file in both the `log-movie.html.twig` file and the `settings-netflix.html.twig` file.
That way we don't have this code duplicated on two separate places

leepeuker: There are sadly multiple places where js code is duplicated, I did handle this part well yet. The star rating logic is in `movie.js` already duplicated if I am not mistaken. 
I would say for now just duplicate it to keep your PR simple and let us clean that up later. You are more than welcome to do that if you are done here. Or even before that in another PR and use this here, so that you can directly use the refactored version here.

JVT038: I may do this in the future.
Because there are multiple ratings on one page (around 91 ratings in my Netflix page), I had to modify the functions to make it work with multiple ratings on one page.
An ID can obviously not be used multiple times for multiple elements, so I initially moved it to the class.
However, I will probably move it to the `data-rating` attribute. You can read more about data attributes [here](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes).
I noticed that the rating attribute is only used by JS to know what the rating is. It's not really used by screenreaders or any stylistic stuff. Just for the rating data.
I therefore find it fitting to use a data attribute instead of a classname. 
And you can probably use the data attribute at other pages as well, as they have the same case, where it's only used for JS.

leepeuker: > However, I will probably move it to the data-rating attribute. You can read more about data attributes [here](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes).

Very good idea, please do that :+1:

JVT038:
It's finally finished and ready for merge!
I have tried to find any and all bugs I could find, but perhaps you can find more.

leepeuker:
I will review and test it as soon as I got time :+1:

JVT038:
The tests are mainly about this:
```
FILE: /mnt/c/xampp/htdocs/Projecten/movary/tests/unit/Api/Imdb/ImdbWebScrapperTest.php
--------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
--------------------------------------------------------------------------------------
 1 | ERROR | [x] End of line character is invalid; expected "\n" but found "\r\n"
--------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
--------------------------------------------------------------------------------------
```

JVT038:
I looked into the tests and they should be fixed now.

leepeuker:
You can ignore [pslam issues](https://psalm.dev/docs/running_psalm/dealing_with_code_issues/), I would prefer adding a Docblock suppression with a short comment why this should be suppressed.

src/Service/Netflix/ImportNetflixActivity.php
```diff
@@ -14,6 +14,7 @@ public function __construct(
 
     /**
      * parseNetflixCSV receives the layout of a Netflix viewing history CSV file and returns them as an associative array
+     * @psalm-suppress PossiblyInvalidArgument
```
leepeuker, line 17: add the comment in the code directly before the place where it happens, otherwise you may suppress valid errors at other places in this method

.gitignore
```diff
@@ -1,7 +1,11 @@
...
 .phpunit.result.cache
-
 .env
 /vendor
 /public/storage
 docker-compose.override.yml
+.phpunit.result.cache
+tmp/*
+.idea/
+vendor/
```
leepeuker, line 10:
```suggestion
```

duplicated

leepeuker:
I am getting an error when uploading a csv
```
[2023-01-26T16:59:00.508829+01:00] movary.EMERGENCY: TMDB API key is probably not set or invalid. {"exception":"[object] (Movary\\Api\\Tmdb\\Exception\\TmdbAuthorizationError(code: 0): TMDB API key is probably not set or invalid. at /app/src/Api/Tmdb/Exception/TmdbAuthorizationError.php:11)
[stacktrace]
#0 /app/src/Api/Tmdb/TmdbClient.php(41): Movary\\Api\\Tmdb\\Exception\\TmdbAuthorizationError::create()
#1 /app/src/Api/Tmdb/TmdbApi.php(54): Movary\\Api\\Tmdb\\TmdbClient->get('/search/movie', Array)
#2 /app/src/HttpController/NetflixController.php(108): Movary\\Api\\Tmdb\\TmdbApi->searchMovie('#amLeben')
#3 [internal function]: Movary\\HttpController\\NetflixController->processNetflixActivity(Object(Movary\\ValueObject\\Http\\Request))
#4 /app/vendor/php-di/invoker/src/Invoker.php(74): call_user_func_array(Array, Array)
#5 /app/vendor/php-di/php-di/src/Container.php(248): Invoker\\Invoker->call(Array, Array)
#6 /app/public/index.php(38): DI\\Container->call(Array, Array)
#7 {main}
"} []
```
I have tried the tmdb sync, that works fine with my api key. Got an idea what is wrong here?

JVT038:
I just checked and TMDB does not have any API scopes, so that can't be it.
The only thing I can think of is that you're using a V4 API key while you're supposed to use a V3 API key

JVT038:
So I added '#amLeben' to my CSV file, and it caused the same error here.
image

public/js/settings-netflix.js
```diff
@@ -0,0 +1,523 @@
...
+    this.getElementsByClassName('tmdbradio')[0].checked = true;
+}
+
+function saveTMDBItem() {
+    let checkedrow = document.querySelector('input.tmdbradio:checked').closest('.tmdbrow');
+    let rowid = document.getElementById('tmdbmodal').dataset.rowid;
+    let targetrow = document.getElementById(rowid); 
+    targetrow.getElementsByClassName('img-fluid')[0].src = checkedrow.getElementsByClassName('img-fluid')[0].src;
+    targetrow.getElementsByTagName('a')[0].href = checkedrow.getElementsByTagName('a')[0].href;
+    targetrow.getElementsByTagName('a')[0].innerText = checkedrow.getElementsByTagName('a')[0].innerText;
```
JVT038, line 457: So for some reason, when searching the columns will change in width. 
I don't know if it is something worth fixing, because it doens't break anything and the search just work like it's supposed to.
It can be experienced as annoying though.

templates/page/settings-netflix.html.twig
```diff
@@ -0,0 +1,105 @@
...
+                <div class="alert alert-secondary" role="alert" style="padding-top: 0.5rem;padding-bottom: 0.5rem">
+                    <b>Download your Netflix history</b>
+                    <br/>
+                    Go to <a href="https://www.netflix.com/settings/viewed">https://www.netflix.com/settings/viewed</a> to view your Netflix history and scroll down to click on the 'download all' link.
+                    <br/>
+                    Upload this CSV file down below.
+                </div>
+                <form class="mb-3">
+                    <label class="form-label" for="file">Upload Your CSV File</label>
+                    <input id="netflixfile" class="form-control" accept="text/csv" name="csvinput" type="file" required="true"/>
```
JVT038, line 34: Should I add that it searches for the name from the Netflix CSV file and not from TMDB?

public/js/settings-netflix.js
leepeuker: I have no issue with with it, I see at as kind of bootstrap default behavior :D

leepeuker:
Found a bug I think:
- upload csv
- click the edit button from one movie in the table
- enter search term "test" and press the search button
=> the table is emptied, all movies are gone
pressing save does not work, logs error:
```
settings-netflix.js:460 Uncaught TypeError: Cannot read properties of null (reading 'getElementsByClassName')
    at saveTMDBItem (settings-netflix.js:460:15)
    at HTMLButtonElement.onclick (netflix:155:92)
saveTMDBItem @ settings-netflix.js:460
onclick @ netflix:155
``

which would be the line
```
targetrow.getElementsByClassName('img-fluid')[0].src = checkedrow.getElementsByClassName('img-fluid')[0].src
```

than when trying to reupload the file (after selecting it again) we get another error

```
settings-netflix.js:49 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'remove')
    at uploadNetflixHistory (settings-netflix.js:49:74)
    at HTMLButtonElement.onclick (netflix:87:114)
```

File: public/js/settings-netflix.js (hunk 9 above)
Comment (JVT038) [reply public/js/settings-netflix.js:462]: Then I'll markt this as resolved

**Comment by JVT038** [comment]
The TMDB search error has been fixed.

**Comment by leepeuker** [comment]
After uploading I cannot select a rating in the table
```
settings-netflix.js:500 Uncaught TypeError: Cannot read properties of undefined (reading 'rating')
    at HTMLElement.setRatingStars (settings-netflix.js:500:34)
```

JVT038:
I hid the Netflix tab by just adding the `d-none` class to it. 
You can merge this code (I haven't found any more bugs) and refactor it if you like; it's your project after all, not mine.

Issue #235, opened 2023-01-17
Add middleware to routes


JVT038:
I think it would be a good improvement to add middlewares to the routes. 
Especially when the controller has to verify whether the user is logged in or not.
For example, this piece of code is repeated across a lot of different controllers:

```php
if ($this->authenticationService->isUserAuthenticated() === false) {
    return Response::createSeeOther('/');
}
```
Wouldn't it be better to handle this in a middleware class? 
I think it would be better to add a middleware to the routes and to check whether the user is logged in, in an auth middleware.
Also add a middleware for the login page that checks if the user is logged out. The login page shouldn't be able to be accessed when the user is logged in.
Furthermore, perhaps create a 'route group' so that one middleware can be specified to apply to multiple routes. This way we don't have to define the same middleware 10 times for 10 different routes, but can just say 'This middleware belongs to these 10 routes' once.

JVT038:
I have looked through the code for a while now, and honestly can't comprehend what's going on, let alone add a middleware.
So could @leepeuker do this one?

leepeuker:
Yes I am aware of this and will do it as soon as it gets to the top of my prio list.

Issue #225, opened 2023-01-11
Find longterm frontend solution 


leepeuker:
The current frontend consists of hacky twig templates and a few thrown together js/css files, which are not so great to maintain.  

Possible solutions for this I see are:
- refactor and improve the twig templates and use something like webpack to manage the js code
- use a framework like vue.js or next.js for the frontend

Right now I would favor building a vue.js proof of concept, mainly because I had the most experience with it in the past. Not sure when I will have the time to look into that more, help appreciated.

perebusquets:
Hi! I've just found this media library and I would be willing to contribute (specially on this topic). 

Mainly I have experience with React but I would also be willing to get hands on with VueJS if need be. Just to make sure we get the right framework/setup for this, I will ask some questions if you don't mind:

1. In the post, you propose NextJS instead of React as an alternative, which is often used for server side rendering. Is this because of some need/functionality? Or just because of the batteries included nature of NextJS?
2. Would you be willing to accept a poc made with React? (in this case I could provide everything set up, including translations, build tools, authentication management, etc pretty quick, as it's what I'm most familiar with).

IMHO, regardless of what framework we use, I think the best path would be a SPA, which gets bundled into a single JS file. This way we could send it in a twig template, and there wouldn't be any need to run an extra docker container or having node installed. It also brings the possibility of easily implementing it as a PWA, having a good mid term solution between only being web based and having a mobile app.

Once a framework is decided, I would also like to suggest that we use a component library for speed of development. With this one, I'm also open to suggestions!

leepeuker:
To be honest I cannot answer most of the question because I do not know the answer without more recherche.

> IMHO, regardless of what framework we use, I think the best path would be a SPA, which gets bundled into a single JS file. This way we could send it in a twig template, and there wouldn't be any need to run an extra docker container or having node installed. It also brings the possibility of easily implementing it as a PWA, having a good mid term solution between only being web based and having a mobile app.

That sounds like a good approach to me, any POC in that direction would be very welcome

perebusquets:
Here's the promised proof of concept! If there are any questions, just ask away!

leepeuker:
Thank you @perebusquets for the POC, that helped me a lot to get a better feeling for the topic. 

I do not think that we will move forward with it right now, I want to focus on making the current setup more stable for version 1.0, but I will keep this open for later versions were we can focus on modernizing the frontend.

perebusquets:
[...]

PR #243, opened 2023-01-22
New Frontend POC

perebusquets:
As discussed in #225 , this is a proof of concept for a next gen frontend for Movary.

For now, this only contains the login page (without any routing and logic). 

In order to avoid having another server in the stack, the way this works, is by placing the built files in the existing `public` folder. Then, whenever a request comes in, it loads a basic `twig` template where it loads the assets (mainly the JS and CSS). 

For now this twig template is only being used in [here](https://github.com/leepeuker/movary/compare/main...perebusquets:movary:frontend-poc#diff-5872e1d861a3c49e922341f78a9f10b9f3c904e779ad40038e0853f6063ca27bR28-R39) in a rather hacky way (with a conditional and a env var). If anyone can provide a better solution for this, where it gets implemented everywhere (maybe in a middleware when it's implemented #235 ?) would be very much appreciated.

Also, right now we're using the `ENV` variable to know if we need to load from the Vite development server (when in development), or serve the built assets in `public/frontend` when in production.

### Implemented features:
- Frontend with Vite for development + SWC for building.
- Template in twig to integrate with the current server.
- Mantine as UI library (right now with the default theme, could be customized in the future).
- Login page with working calls to the server as an example for other developers to start working on it.
- ReactQuery for managing the state with the server.
- React Router.
- i18next, with a custom backend to use Vite to build and serve the translation files.

### Pending requirements from the backend before we move to this frontend:
- Add a way in the server to render the new twig template based on an env var (this way, we can transition more smoothly into the new frontend and give the users an opportunity to try out the new frontend without removing the old frontend in case we miss some functionality).
- Do endpoints (REST, GraphQL,...) that accept/return the requests/responses in the body, instead of URL encoded GET and POST requests. The problem now is that after executing the login, the server does a redirect and returns a twig template, which makes impossible for the new frontend to know which path it should go to to show the dashboard.

### To test locally:
Simply start docker, and make sure that the `USER_ID` in the `.env` is the user on your local machine. Also add `NEW_FRONTEND=1` to load the new template.


JVT038:
So I honestly have no experience with React, so I can't really judge this.
What you've written in this issue sounds promising, but @leePeuker would need to look at this as well.
Either way, I'm looking forward for the implementation of this!

JVT038:
BTW, assuming this replaces all the hardcoded strings with the translations, in the html templates, can you also use this to replace the hardcoded strings in the JS /PHP files?

leepeuker:
@JVT038 I talked about this POC in the github [issue](https://github.com/leepeuker/movary/issues/225)

perebusquets:
Hi @JVT038 ! Thank you for showing your interest in this. 
To respond to your last question, this doesn't replace the hardcoded strings in the current templates. 
This implementation of translations just works with the new frontend, where by calling the funcion `t` from the library `i18next` it replaces the given key for the translation. 
Since this library only runs in JS, at the time of rendering, it would be impossible to use after PHP has rendered the templates.

PR #250, opened 2023-01-27
Add backdrop image to movie page

leepeuker:


JVT038:
I think this design looks better and cooler than the current, so I would merge this.
But when the movie backdrop is the same colour as the movie details next to the poster, then the movie details are not very readable anymore.

Example:
image

The genres are "History, Action, War" but it's not very clear.

JVT038:
I would be in favour of something like Trakt.tv, where they have a dark, transparent background to properly ensure that the text is readable, despite the backdrop.

image

leepeuker:
> I think this design looks better and cooler than the current, so I would merge this. But when the movie backdrop is the same colour as the movie details next to the poster, then the movie details are not very readable anymore.
> 
> Example: image
> 
> The genres are "History, Action, War" but it's not very clear.

My thoughts:
- we must improve the readability of the text
- I do not like the the user select being included into the backdrop I think :thinking:
- I think the backdrop image should not stop in the middle of the logo, that looks weird, maybe move it down and let it start at the name and stop at the end of the ratings/start of the buttons? I do not really now how it could be improved, I just know that it looks not right to me :D

PR #256, opened 2023-01-28
Netflix import cleanup

leepeuker:
Todo:
- ~~delete button for movies in table~~ (created [ticket](https://github.com/leepeuker/movary/issues/258))

Notable changes:
- search results in tmdb modal are cleared when modal is closed
- rows of movies without tmdb matches have now a yellow background to warn the user of the missing match
- watch dates are only imported if the movie does not already have that watch date (the original way just counted the plays for the date up, that created the risk of watch date duplications)
- displaying watch dates in table with the user chosen format (removed hardcoded format)


leepeuker:
@JVT038 In case your interested, I will do my refactoring/improvements on the Netflix import here

src/HttpController/NetflixController.php
```diff
@@ -23,114 +23,109 @@ public function __construct(
...
+    }
 
     /**
-     * importNetflixActivity receives a CSV file with all the Netflix activity history and tries to process this. 
+     * Receives a CSV file with all the Netflix activity history and tries to process this.
      * It filters the movies out with regex patterns, and then compiles an array of all the movie items.
-     *
-     * @param Request $request the CSV file containing the Netflix data
-     * @return Response HTTP response with either an error code or JSON object containing the TMDB results 
      */
```
JVT038, line 36: Why have you removed these things?
I personally find them very useful, as they provide the necessary information for my IDE to properly show and describe a function.

src/HttpController/NetflixController.php
```diff
@@ -23,114 +23,109 @@ public function __construct(
...
         }
+
         $csv = $files['netflixviewactivity'];
-        if($csv['size'] == 0) {
+        if ($csv['size'] == 0) {
             return Response::createBadRequest();
         }
-        // finfo_open is way more reliable to detect MIME type than 'type' from the $_FILES variable
-        // It does however require the magic module, which may be a pain to set up.
-        // https://www.php.net/manual/en/function.finfo-open.php
```
JVT038, line 54: BTW, do you think simply checking the file extension is sufficient to verify that it's actually a CSV file?
Or would you rather use the Magic module, that actually inspects the file data itself and can check the MIME type very accurately?

JVT038 (changes requested):

src/HttpController/NetflixController.php
JVT038, line 69: Why have you removed these things?
I personally find them very useful, as they provide the necessary information for my IDE to properly show and describe a function.
They can additionally be useful for other developers who want to contribute. 
Adding comments like these can be really useful and helps to describe a method / function.

src/HttpController/NetflixController.php
JVT038, line 54: BTW, do you think simply checking the file extension is sufficient to verify that it's actually a CSV file?
Or would you rather use the Magic module, that actually inspects the file data itself and can check the MIME type very accurately?

src/HttpController/NetflixController.php
leepeuker: Probably because of habit, I am not a friend of comments most of the time. These docblock comments do not provide much useful information you cannot (or should be able to) get from the methods signature itself and can be downright wrong if not maintained well (code changed but comments did not and things like that).

For example:

> @param Request $request the CSV file containing the Netflix data

The type should be clear by the methods signature and what the parameter contain can be made clear through naming (class name is `NetflixController` and parameter `Request $csvImportRequest`).

> @return Response HTTP response with either an error code or JSON object containing the TMDB results

Same here, return type is defined via methods signature. That a http response can have an error code should be clear by the nature of http responses and that the successfully response can be made clear through method naming, e.g. `provideTmdbMatchesForNetflixCsv`.  

In my experience comments lead developers to be lazy, why find good namings and improve code readability when you can write a comment to explain an unreadable mess, at least that is what comments are inherently are supporting even if you do not want it.

This does not count for all comments and all situations of course. There are totally valid comments explaining difficult code or giving a high level overview about an entry method to a complex process and you are of course correct that it can be useful for contributors, but to be honest right now I would prefer less contributors than to work with docblocks causing the issues I see mentioned at the beginning.

src/HttpController/NetflixController.php
leepeuker: Yeah I think it is fine, I see no reason to be super correct here, I not think it should impact usability or open us to any attack vectors.

src/HttpController/NetflixController.php
JVT038: I get your point, however the parameters occasionally require a comprehensive and detailed description, which can't be put in the naming. Or you would get absurdly long variable names, but that would decrease readability.

And I also just think it looks cool for my IDE, but this argument is probably just a big fallacy, hahaha.

Anyway, I'll keep this in mind and remove the `@return` and `@param` tags then. I'll probably keep the description and summaries in the docbloc.

JVT038 (approved):

src/HttpController/NetflixController.php
JVT038: Yeah, you're right; the code only tries to read the file data and if it isn't a CSV, it'll probably throw an error.

src/HttpController/NetflixController.php
leepeuker: > however the parameters occasionally require a comprehensive and detailed description, which can't be put in the naming. Or you would get absurdly long variable names, but that would decrease readability

Then I have no issue with a comment

> And I also just think it looks cool for my IDE, but this argument is probably just a big fallacy, hahaha.

No I totally get you :D 

You can definitely keep comments you find necessary because you cannot express this information via code, I have no issue with that. Just keep in mind comments could/should be removed when the code was for example refactored to be better understandable (hopefully :see_no_evil: ).

JVT038 (commented):
Why have you moved the logic to `ValueObject`?
What should `ValueObject` even contain?

leepeuker:
> Why have you moved the logic to `ValueObject`? What should `ValueObject` even contain?

I have renamed the namespace to `Dto`, these classes are not value objects (if you want to know more about value objects I can recommend to read [this](https://martinfowler.com/bliki/ValueObject.html)).

I do not like to pass arrays around between classes. Arrays are inherently type unsafe because of their dynamic nature and difficult to maintain (e.g. when the structure changes you have a lot of manual work to do for fixing all usages, that is much easier having an object woth getter methods for the data).

README.md
```diff
@@ -346,3 +347,9 @@ The application should be up-to-date and running locally now.
 
 - Report bugs or request features via github [issues](https://github.com/leepeuker/movary/issues)
 - Ask questions or discuss movary related topics in the [official subreddit](https://www.reddit.com/r/movary/)
+
+
+## Contributors
+
+* [@leepeuker](https://github.com/leepeuker) as Lee Peuker
+* [@JVT038](https://github.com/JVT038) as JVT038
```
leepeuker, line 355: @JVT038 If you do not want to mentioned I can remove this

leepeuker:
@JVT038 So, I think I am done here. Please test it again to make sure I broke nothing, it seems good to me. Checkout the first comment of the PR to see what I changed besides the refactoring.

The only issue I see is that we should maybe change the actual import to be processed as a background job and not during the http request to prevent issues like timeouts. I have tested a import with ~300 movies and it went reasonable well. Not sure about bigger histories

leepeuker:
I have removed merged only the refactoring and changes I mentioned at the top. We have to fix https://github.com/leepeuker/movary/issues/259 before making it officially available

PR #257, opened 2023-01-28
Added registration with new environment variable

JVT038:
Done: 
- [x] Moved create user logic to the
- [x] CreateUserController
- [x] Changed title of create user page to 'create user'
- [x] Added extra route for the create user page

So I have added the posibility to set an environment variable called `REGISTRATION` to 'true' or 'false'. If it's true, the login page will contain a new link below the 'login' button where the user will be redirected to a new create user page. This create user page is rendered from the CreateUserController; it makes a lot more sense to me to move the rendering logic of this page from the LandingPageController to the CreateUserController.


JVT038 (commented):
Wow, this actually ran all tests without any errors, so that's good as well.

JVT038:
I think the registration should be changed to an AJAX / XHR request.
Because when the user is currently filling in the form and the user has done something wrong (e.g. the passwords aren't the same), the entire form is cleared, because the window has to be reloaded.
When using an AJAX / XHR request, the window won't reload and the form won't be cleared on any errors.

JVT038:
@leepeuker I think it's done, could you check the PR?

PR #263, opened 2023-01-30
Add custom date format definition to Netflix import

leepeuker:
TODO:
- [x] Disable `Upload File` button until a file was chosen and the date format set
- [x] Add the date format value to the upload request
- [x] ~~Validate the submitted date format~~ (not sure how) 
- [x] Use the submitted date format when processing the CSV

### DIFF
--- a/public/js/settings-netflix.js
+++ b/public/js/settings-netflix.js
@@ -29,43 +29,41 @@ async function importNetflixHistory() {
         body: importDataAsJson
     }).then(response => {
         if (!response.ok) {
-            processError(response.status);
+            console.log(response)
+            setImportAlertError();
 
             return
         }
-        let td = document.createElement('td');
-        let tr = document.createElement('tr');
 
-        td.colSpan = 4;
-        td.innerHTML = '<p class="text-success">The data has succesfully been imported!</p>';
-        tr.append(td);
-        document.getElementById('netflixTableBody').append(tr);
-        document.getElementById('netflixTableBody').getElementsByTagName('tr')[0].remove();
+        setAlert('importAlert', 'The data has been successfully imported!', 'success')
+        setDefaultTableContent()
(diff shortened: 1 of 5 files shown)


leepeuker:
@JVT038 This is my POC for the UI, what do you think?

JVT038:
Looks good.
Now we'll just have to assume that the users aren't stupid and can properly understand the examples and apply them for their own use case.

JVT038:
Also, add a verification on the back end that checks whether the input format is valid or not.

leepeuker:
Users are almost always stupid, but I see no better way :sweat_smile:

JVT038:
BTW, can't you use this some concept to allow custom date format in the account settings?

leepeuker:
> BTW, can't you use this same concept to allow custom date format in the account settings?

That is another situation imo, entering something unexpected/stupid in the account settings format could potentially break the app at a lot of places vs here in the import, where it only affects a small module of the app.

leepeuker:
@JVT038 I think I am done here. Could you test it again? I have changed a few additional things like the error/success messages (using bootstrap alerts) and the importer description text.

JVT038:
>  Validate the submitted date format (not sure how)

The submitted date is automatically validated with the `createFromStringAndFormat()` method, because it checks `$dateTime === false` and if it's false, then the date couldn't be formatted according to the input format.
When it's false, an exception is automatically thrown and the user sees an error, so this is handled already.

leepeuker:
True, I thought there was maybe some way to check if a string is a valid date format without directly trying to use it, but could not find something

JVT038:
You can probably use regex for this

leepeuker:
Yeah but that is not worth the effort, that would be a pretty complicated regex :D

JVT038:
Agreed lol

PR #267, opened 2023-01-31
Fixed page navigation

JVT038:


leepeuker:
Awesome, that fixed a lot.

Still buggy:
- when the filter `Not found` is active at one item is displayed the pagination looks correct. But when clicking on the page `1` button the table is emptied and nothing is displayed anymore
- when the filter `Not found` is active and some search term is added, the pagination resets to the original filterless pagination (it jumped from 1 page with filter (correct) to 4 pages (wrong, pages without `Not found` filter) when entering any character into search for me)

JVT038:
I had decided that the pagination would be 'disabled' when the filter `Not found` is enabled.
My reason is because the pagination system is already very complicated and adding pagination for the items that aren't found would be very difficult for me to implement.
I'll just completely disable the pagination when the `Not found` filter is enabled.

leepeuker:
> I had decided that the pagination would be 'disabled' when the filter Not found is enabled.

Fine by me :+1: 

Sadly the second issue I reported is still happening:

When the filter `Not found` is active and some search term is entered, the table is emptied and the pagination matches the filterless search state. You have to unset the filter to see something in the table again.
Maybe we can just disable the search input when the `Not found` filter is active similar to the pagination?

JVT038:
>  Maybe we can just disable the search input when the Not found filter is active similar to the pagination?

Yeah, let's do that.
If the user wants to search with the `Not found` filter, they can just use the built-in search function in their browser (CTRL + F)

PR #268, opened 2023-01-31
Add note symbol to watch dates

leepeuker:


templates/page/movie.html.twig
```diff
@@ -126,13 +126,14 @@
...
                     {% for watchDate in watchDates|reverse %}
-                        <li class="list-group-item {% if currentUsername == routeUsername %}list-group-item-action{% endif %}"
+                        <li class="list-group-item {% if currentUsername == routeUsername %}list-group-item-action{% endif %} clearfix"
                             style="text-align: center;{% if currentUsername == routeUsername %}cursor: pointer{% endif %}"
                             {% if currentUsername == routeUsername %}onclick="loadWatchDateModal(this)"{% endif %}
                             data-watch-date="{{ watchDate.watched_at|date(dateFormatPhp) }}"
                             data-plays="{{ watchDate.plays }}"
                             data-comment="{{ watchDate.comment }}">
                             {{ watchDate.watched_at|date(dateFormatPhp) }} {% if watchDate.plays > 1 %}({{ watchDate.plays }}x){% endif %}
+                            {% if watchDate.comment != '' %}<i class="bi bi-chat-square-text float-end">{% endif %}</i>
```
leepeuker, line 136: this breaks the UI atm... the note icon is on the right site, but the text is not centered anymore but is too far on the left because of the added icon.

I am sure there is a easy fix like using the `clearfix` in line 129, but it does not work :(

leepeuker: @JVT038 Do you maybe have an idea, I cannot find a fix for this...

leepeuker: okay forgot it, found a solution :D

PR #269, opened 2023-02-01
Disable the input fields during 'Not found' filter

JVT038:
This disables the input fields instead of hiding them.

### DIFF
--- a/public/js/settings-netflix.js
+++ b/public/js/settings-netflix.js
@@ -167,8 +167,10 @@ function updateTable() {
         return;
     }
 
-    document.querySelector('label[for="amountToShowInput"]').classList.remove('d-none');
-    document.getElementById('amountToShowInput').classList.remove('d-none');
+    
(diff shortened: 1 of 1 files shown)


JVT038:
Unrelated to this PR specifically, but could you start ending javascript statements with a semicolon?
It's generally recommended to do this, as it prevents any parsing mistakes by the browser from occuring. [Stackoverflow thread](https://stackoverflow.com/questions/444080/do-you-recommend-using-semicolons-after-every-statement-in-javascript)

leepeuker:
> Unrelated to this PR specifically, but could you start ending javascript statements with a semicolon? It's generally recommended to do this, as it prevents any parsing mistakes by the browser from occuring. [Stackoverflow thread](https://stackoverflow.com/questions/444080/do-you-recommend-using-semicolons-after-every-statement-in-javascript)

Yeah probably a good idea. I am using phpstorm default settings for js formatting which removes semicolons. I can update this

PR #274, opened 2023-02-03
Import Plex watched history and basic implementation of the Plex API

JVT038:
This is the start of importing Plex watched history.

I've implemented the very basic of the API, but I've already hit a weird problem that I haven't solved.

The authentication flow is supposed to be this: 

1. User visits `/settings/plex` 
2. The backend will check if an access token is stored in the database and if there is one, it will be verified if the token actually works.
3. If the token works, the 'login with plex' button is disabled and this is where the workflow ends (for now). But if it doesn't work, it won't be disabled and this continues to step 4.
4. The backend will generate an authentication URL from Plex, by sending them a request for a PIN. Plex will respond with a pin and a token. This token is stored in the database and an authentication URL is constructed from the received plex data.
5. The user clicks on the log in button and will be redirected to the authentication URL.
6. The user logs in and authorizes Movary access to their account.
~~7. This is where it goes wrong. For some reason, when I authorize Movary, it will say that the request has failed. In the networks tab, the response has a message that says "This action is not available for this device.". And because of this, each time Movary verifies the access token, the response will be error 401, because it can't be properly authorized.~~ This issue seems to be fixed.

~~There was a forum post about this, [here](https://forums.plex.tv/t/plex-oauth-not-working-with-tautulli-ombi-etc/433945) and the problem was internal. A Plex employee fixed it, but we can't fix this.
I don't know how to solve this problem, if it is solvable. I made a post on their forum, so hopefully an employee will respond and fix it, like the last time.
A 'temporary' fix is by changing the product name to 'Plex Web', but we shouldn't spoof our application by naming them 'Plex Web'. That shit feels illegal to do, and it confuses the user when it sees that an application called 'Plex Web' is requesting access.~~

EDIT: 7. However, there is another issue. The response is error 401, even when I've granted Movary access to my account, so I'll be investigating this.

I don't know if the issue is going to be fixed. If it isn't, then we'll have to find a workaround.
I have various things in mind:

~~1. We literally sign in with the password and username into the central authentication API of Plex. That feels unsafe, because we would have to plainly store their password in the database. Using their official OAuth and getting an access token is way safer. Also, I don't know if this will work if the user has enabled Two Factor Authentication.
2. Connect to the local Plex server instead of the central API. This should probably fix the issue, but will be a bit more complicated, I think.
3. Log in with a PIN code. In this case, the app will request a pin code from Plex, the user will enter this pin code in their account, which authorizes Movary and Movary will receive an access token. But this would probably require a web socket connection, because Movary would need to poll the Plex server every second to check whether the pin code has been entered or not.
4. Ask the user to enter their own `X-Plex-Token`. This can be manually retrieved with [this method](https://www.plexopedia.com/plex-media-server/general/plex-token/), but it is a bit more advanced and I don't know if users without privileges are able to do this. If a user isn't admin or anything like that, can they get the `X-Plex-Token`?~~

~~Do you have any ideas? Here are some useful URLs
https://github.com/Arcanemagus/plex-api/wiki
https://forums.plex.tv/t/authenticating-with-plex/609370
And my post on their forum: https://forums.plex.tv/t/status-403-with-error-1068-and-message-this-action-is-not-available-for-this-device/829391~~

EDIT: The authentication issue has been fixed.


JVT038:
Plex screenshot after login:
image

Plex network response:

```json
{
	"code": 1068,
	"message": "This action is not available for this device.",
	"status": 403
}
```
Plex JSON response in Movary:
```json
{
    "errors": [
        {
            "code": 1001,
            "message": "User could not be authenticated",
            "status": 401
        }
    ]
}
```

JVT038:
So I have fixed error 1068, but I have now encountered something very weird.
Whenever I try to login via Movary, it will succeed, but checking the pin always returns HTTP error 404, with the message "Code not found or expired".
When I try to re-enact all HTTP requests with curl in my CLI, this doesn't happen.
Will keep investigating.

JVT038:
OK, so I don't know how and/or why, but it suddenly works and I didn't do anything?? I guess the Plex API works in mysterious ways...

JVT038:
@leepeuker So I've fixed authentication, and I've started experimenting with the Plex API, to discover its quirks.

Well, I've found multiple quirks so far.
First off, the Plex API for authentication is the official Plex URL; `https://plex.tv/api/v2`. The Plex API for the server data to retrieve info about the libraries, movies, etc. is the local server url. (something like `http://localhost:32400` or `plex.privatedomain.com` for reverse proxies). Authentication is impossible via the local Plex API, only via the official one.
Second, while the official Plex API returns JSON data, the local Plex API (necessary for the movie info and stuff) only returns XML data.

So because there are two major differences between the two APIs (The returned data type is different and the base url is different), I'm considering to make two different plex clients. There is currently one file `src\Api\Plex\PlexClient.php`, that submits the HTTP requests, but I think changing the name to `PlexAuthClient` or something like that, because that file would only handle the HTTP requests required for authentication. Then I'd make a new file called `PlexLocalClient` to handle the local HTTP requests, which are necessary for everything except authentication.

Another thing that I've been wondering, is the temporary pin code.
To authenticate, a temporary pin code is requested and needs to be added to the HTTP request to get an access token. This pin code is currently stored in the database (`user` table), but it's only valid for ~15min. Should we remove the column from the database, and instead store the temporary pin code in the cookies? There we can actually set an expiry date as well.

.env.production.example
```diff
@@ -17,5 +17,10 @@ DATABASE_MYSQL_CHARSET=utf8mb4
 TMDB_API_KEY=
 TMDB_ENABLE_IMAGE_CACHING=1
 
+# Plex
+# The identifier can be any random string. It can be generated with the command 'openssl rand -base64 32'
+# This is used to identify the Plex app
+PLEX_IDENTIFIER=
```
leepeuker, line 23: should this be a global setting or unique per user?

JVT038: The identifier should be unique per client, and I'm assuming we'll be using the same client for each user right?
Though I have seen some other projects that use an unique identifier for each user.

I ultimately don't think it really matters a lot.

src/Api/Plex/dto/PlexAccessToken.php
```diff
@@ -0,0 +1,21 @@
+<?php declare(strict_types=1);
+
+namespace Movary\Api\Plex\Dto;
```
leepeuker, line 3: Please change `dto` to `Dto` in the file path

src/Api/Plex/PlexLocalServerClient.php
```diff
@@ -0,0 +1,90 @@
...
+        $options = [
+            'form_params' => $data,
+            'headers' => $httpHeaders
+        ];
+        $response = @$this->httpClient->request('get', $url, $options);
+        $statusCode = $response->getStatusCode();
+        match(true) {
+            $statusCode === 401 => throw PlexAuthenticationError::create(),
+            $statusCode === 404 => throw PlexNotFoundError::create($url),
+            $statusCode !== 200 && $statusCode !== 201 && $statusCode !== 204 => throw new RuntimeException('Plex API error. Response status code: '. $statusCode),
```
leepeuker, line 57:
```suggestion
            $statusCode === 200 || $statusCode == 201 && $statusCode !== 204 => throw new RuntimeException('Plex API error. Response status code: '. $statusCode),
```

src/Api/Plex/PlexLocalServerClient.php
```diff
@@ -0,0 +1,90 @@
...
+            default => true
```
leepeuker, line 58:
```suggestion
            $statusCode === 200 || $statusCode === 201 || $statusCode || 204 => true,
            $statusCode === 401 => throw PlexAuthenticationError::create(),
            $statusCode === 404 => throw PlexNotFoundError::create($url),
            default => throw new RuntimeException('Plex API error. Response status code: '. $statusCode),
```

this seems easier to understand to read to me

src/Api/Plex/PlexLocalServerClient.php
```diff
@@ -0,0 +1,90 @@
...
+        $url = $this->plexServerUrl->getPlexServerUrl() . $relativeUrl;
+        $postData = array_merge($this->defaultPostAndGetData, $customPostData);
+        $httpHeaders = array_merge($this->defaultPostAndGetHeaders, $customPostHeaders);
+        $options = [
+            'form_params' => $postData,
+            'headers' => $httpHeaders
+        ];
+        $response = @$this->httpClient->request('post', $url, $options);
+        $statusCode = $response->getStatusCode();
+        match(true) {
```
leepeuker, line 81: see https://github.com/leepeuker/movary/pull/274/files#r1100529166

src/Api/Plex/PlexLocalServerClient.php
```diff
@@ -0,0 +1,90 @@
...
+use Movary\Api\Plex\Exception\PlexNotFoundError;
+use RuntimeException;
+
+class PlexLocalServerClient
+{
+    private const APP_NAME = 'Movary';
+    private $defaultPostAndGetData;
+    private $defaultPostAndGetHeaders;
+
+    public function __construct(private readonly httpClient $httpClient, private readonly Config $config, private readonly PlexServerUrl $plexServerUrl)
```
leepeuker, line 19: I would prefer to not inject `Config` directly but only the needed values. it should not be responsibility of this class to work with `config`, let this work the done by the factory

src/Api/Plex/PlexLocalServerClient.php
```diff
@@ -0,0 +1,90 @@
...
+        $this->defaultPostAndGetData = [
+            'X-Plex-Client-Identifier' => $this->config->getAsString('PLEX_IDENTIFIER'),
+            'X-Plex-Product' => self::APP_NAME,
+            'X-Plex-Product-Version' => $this->config->getAsString('APPLICATION_VERSION'),
+            'X-Plex-Platform' => php_uname('s'),
+            'X-Plex-Platform-Version' => php_uname('v'),
+            'X-Plex-Provides' => 'Controller',
+            'strong' => 'true'
+        ];
+        $this->defaultPostAndGetHeaders = [
```
leepeuker, line 30: this can be a constant, static value

db/migrations/mysql/20230202165432_add_plex_access_and_id_tokens.php
```diff
@@ -0,0 +1,28 @@
+<?php declare(strict_types=1);
+
+use Phinx\Migration\AbstractMigration;
+
+final class AddPlexAccessAndIdTokens extends AbstractMigration
```
leepeuker, line 5: I would prefer if you merge the migration files for the final pr, easier to manage smaller amounts of total migrations imo :thinking:

leepeuker, line 5: Reminder comment that sqlite migration is missing :sweat_smile:

src/Api/Plex/PlexLocalServerClient.php
leepeuker: you are repeating this match statement a lot, maybe we should refactor this into one place (maybe a trait to keep it simple?)

src/Api/Plex/PlexLocalServerClient.php
JVT038: Should I create a Dto for the Plex Identifier and for the application version and inject that instead?

leepeuker: Hm, I think for the start I would be fine with just strings here atm, if the user enters an invalid structured string the api will return an auth error, which we handle with `PlexAuthenticationError`

JVT038:
So I'm trying to figure out which Plex items are completely 100% watched and which aren't.
Apparently, the `/status/sessions/history/all` adds items whenever they're played. Even when an item is interrupted halfway through and not completely finished, it's included in that response. So it's unreliable to check whether an item has been watched or not.
I have yet to find a reliable endpoint that really tells whether the item has been watched and when it has been watched.

Will keep investigating.

public/js/settings-plex.js
```diff
@@ -77,3 +77,59 @@ async function deletePlexWebhookIdRequest() {
...
+
+async function createSpinner() {
+    let div = document.createElement('div');
+    let span = document.createElement('span');
+    div.className = 'spinner-border';
+    span.className = 'visually-hidden';
+    span.innerText = 'Loading...';
+    div.append(span);
+    document.getElementById('plexImportHistoryLog').append(div);
+}
```
JVT038, line 125: I think we should refactor this, because I literally just copied the Netflix spinner into here.
This will have to be refactored together with the rating star things.

public/js/settings-plex.js
```diff
@@ -77,3 +77,59 @@ async function deletePlexWebhookIdRequest() {
...
+
+function enable(el) {
+    el.classList.remove('disabled');
+    el.removeAttribute('disabled');
+}
+
+function disable(el) {
+    el.classList.add('disabled');
+    el.setAttribute('disabled', '');
+}
```
JVT038, line 135: I think we should refactor this, because I literally copied it from the Netflix javascript part.
This can be refactored together with the rating stars

JVT038:
Also, not all Plex items have a TMDB ID linked to them.
Personally, I have disabled the whole online Plex metadata part. So this means that I don't have a TMDB ID or IMDB ID linked to any of my items.
Additionally, some items Plex literally just can't find on TMDB for whatever reason.
To fix this, I'll probably just make a similar workflow to the Netflix import; generate a table with all the Plex items, try to match them with data from TMDB, and have the user manually edit them if they're incorrect and/or can't be found.
Any thoughts on this?

leepeuker:
> Any thoughts on this?

Hm, I would be fine if we just only import movies with existing tmdb ids, keeps things much easier on our site and would be enough for most people. but if you want to implement his I agree, the netflix way would make sense. than we maybe should refactor the whole table logic into a separate component, which we can use on any number of import pages :thinking: otherwise we would duplicate to much (like with the stars)

JVT038:
> Maybe let us make a version 1 of the importer which just imports movies with tmdb ids (the core logic) and when this is done and merged we work on version 2 with the tmdb match table?

Sounds good to me.

JVT038:
So, another update.
I haven't really written any code, because the Plex API is way more complicated than expected.

`/status/sessions/history/all` returns the data of _all_ users, not only one. 
I thought that would be it. But no, unfortunately, when testing other users, Plex will return error 401. Apparently, the endpoint `/status/sessions/history/all` is restricted to admin users only. Therefore, other users in the same household or friends who are watching on the server of another admin, can't view their own watch history. 
Yes. You literally can't view your _own_ watch history if you're not an admin on the server.
Which seems a bit logical, considering the endpoint returns the data for _all _ users. Otherwise, a random user on the Plex server can see what another user on the same server is watching and has watched in the past.

And additionally, I still haven't solved the problem that I don't know when an item has been fully watched or not.

JVT038:
Is this the end? Should I just stop developing this PR? 
I don't know, I'm a bit disappointed to say the least.

JVT038:
Maybe we should use [Tautulli](https://tautulli.com/) instead? 
They at least have a proper [API](https://github.com/Tautulli/Tautulli/wiki/Tautulli-API-Reference) that we can use to fetch the user [history](https://github.com/Tautulli/Tautulli/wiki/Tautulli-API-Reference#get_history)

leepeuker:
I am sorry to hear about the problems you have encountered and can totally feel your frustration!
Before you get burned out with Plex I agree it would probably best to stop and maybe work on other things.

JVT038:
So a kind Plex employee told me the following:

> 1 - Not with that api. That api only tracks when something is played, not how much. Alternatively you could call up the library and filter for unwatched=0, which will return all watched items for that library. You’d have to make this request for each library.
2 - Normal users don’t have access to the history, that is admin only. The alternative above will work with other users.
The alternative above will include items that were “marked as watched” too. I’m not aware of an endpoint that will give a list of things actually watched. If that’s what you want, I guess you could use the history and cross reference with this alternative list.

I tested this and it works.
One small caveat however: It will only show the watched stuff once. So for example, if I watch the same movie three times on three separate occasions, this method will only provide you with _one_ time that you have watched the movie.
The `lastViewedAt` attribute will provide us with a UNIX timestamp of the time it was watched. I think we can presume that whatever date that attribute contains, is the date that the item was watched.


So I think I'll continue working on it, as an employee just saved this PR lol.

leepeuker:
I think that sounds like the other plex importer I mentioned works like (see [here](https://github.com/ArabCoders/watchstate/blob/master/src/Backends/Plex/Action/Import.php)), he is looping through all libarary sections. I used this to sing my history to jellyfin and it was able to get the watch history for a given user from plex, so there definitly is a way

JVT038:
> Hm, that sucks but probably better than nothing 🤔

Yup. I recently discovered Netflix does the same. If I watched the same episode twice, the episode will only appear once in the Netflix CSV file.
The watched date will be the most recent date that I've watched the episode / movie.

JVT038:
I think I have finished the basics of the Plex import now.

Movary can now fetch all the libraries, loop through the libraries and find out which library contains movies.
When a library contains movies, movary will fetch all the movies from the library and check which ones have a status of `unwatched=0`. 
The movies who are not unwatched (so they've been watched), are then checked for a TMDB ID and an IMDB ID.
If a movie does not have a TMDB ID, movary will add that movie to a list and continue looping without registering this movie as watched in the database.
If a movie does have a TMDB ID, Movary will add that movie to the movie table if it is not there yet, check if the movie has been watched on the same date already and if it has not been watched on the same date, Movary will finally add the movie to `movie_user_watch_dates` table.

JVT038:
@leepeuker I'll fix the other tests in the upcoming days, but what do you think of it?

leepeuker:
I will look into it in the next days, I do not have much time right now, sorry!

db/migrations/mysql/20230202165432_add_plex_columns.php
```diff
@@ -0,0 +1,32 @@
...
+            ALTER TABLE user DROP COLUMN plex_client_id;
+            ALTER TABLE user DROP COLUMN plex_client_temporary_code;
+            ALTER TABLE user DROP COLUMN plex_access_token;
+            ALTER TABLE user DROP COLUMN plex_account_id;
+            ALTER TABLE user DROP COLUMN plex_server_url;
+            SQL,
+        );
+    }
+
+    public function up() : void
```
leepeuker, line 20: I would prefer a separate table for the plex import data to group the responsibilities a bit more :thinking: and it makes the sqlite migration easier :D

.env.production.example
```diff
@@ -17,5 +17,10 @@ DATABASE_MYSQL_CHARSET=utf8mb4
 TMDB_API_KEY=
 TMDB_ENABLE_IMAGE_CACHING=1
 
+# Plex
```
leepeuker, line 20: add to .env.development.example

src/Api/Plex/PlexTvClient.php
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * @throws PlexNotFoundError
+     * @throws PlexAuthenticationError
+     * @throws PlexNoClientIdentifier
+     * @throws RuntimeException
+     * @psalm-suppress InvalidReturnType
+     */
+    public function sendGetRequest(string $relativeUrl, array $customGetData = [], array $customGetHeaders = [], ?string $customBaseUrl = null) : array
+    {
+        if ($this->config->getAsString('PLEX_IDENTIFIER', '') === '') {
```
leepeuker, line 48: inject the identifier directly into the class instead of `Config`

src/Api/Plex/dto/PlexAccessToken.php
leepeuker, line 1: you have to fix the file path, `dto` to `Dto`, otherwise it does not work

src/Api/Plex/dto/PlexItem.php
```diff
@@ -0,0 +1,66 @@
...
+
+class PlexItem
+{
+    public function __construct(
+        private readonly int $itemId,
+        private readonly string $type,
+        private readonly string $title,
+        private readonly ?PersonalRating $userRating,
+        private readonly ?int $tmdbId,
+        private readonly ?string $imdbId,
```
leepeuker, line 17: I do not think we need the imdbId? :thinking: we import via tmdbId and get the imdbId from there

db/migrations/mysql/20230202165432_add_plex_columns.php
JVT038: Could you elaborate on this? What do you mean with grouping the responsibilities and why do you think this is important?
The sqlite migration doesn't necessarily need to be easier imo, it's clear to me what the queries do.

src/Api/Plex/dto/PlexItem.php
JVT038: The Plex API returns various IDs from various services. 99% of the time the TMDB ID is present, just like the IMDB ID. (This is all dependent on the user's metadata settings though, if I disable the TMDB metadata agent, no TMDB ID will be returned by the API)
I think it's better to try and fetch the IMDB ID from Plex, and resort to TMDB if no IMDB ID has been found. This way, we can reduce the amount of API calls being made to TMDB, and reduce the network / bandwith / wifi thingy use.

public/js/settings-integration-plex.js
```diff
@@ -77,3 +77,59 @@ async function deletePlexWebhookIdRequest() {
         throw new Error(`HTTP error! status: ${response.status}`)
     }
 }
+
+async function sendPlexImportRequest() {
+    disable(document.getElementById('triggerPlexImportBtn'));
+    await  createSpinner();
+    await fetch('/settings/plex/importplexhistory', {
```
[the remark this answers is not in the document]
JVT038: I honestly have no idea, I'll test it later with ~100 movies when I have time.

src/Api/Plex/dto/PlexItem.php
leepeuker: Hm, it seems like I did not check the implementation in detail. There should be always a request to tmdb api when importing a movie, in which case we always can take the imdb id. Having tmdb api as the one point of truth for the meta data makes the managing of it much easier, at least at the moment

PR #275, opened 2023-02-04
Replace third party movary logo with self made logo

leepeuker:


leepeuker:
Version 1

Not really feeling yet, still working on it

PR #284, opened 2023-02-15
Add dark mode

leepeuker:


leepeuker:
@JVT038 I would appreciate some feedback, I think I am done with the basic implementation of the dark mode. The theme buttons are in the menu dropdown. What do you think about it?

JVT038:
I briefly checked this out, and it looks good!

One thing I'm unsure of, is the dark mode button.
I would personally prefer a [switch](https://getbootstrap.com/docs/5.2/forms/checks-radios/#switches) instead of two different buttons with a sun / moon icon.

In another project of mine, I had dark mode in the navbar like this:
image

leepeuker:
Oh yes, a switch is a great idea, I will try this!

PR #289, opened 2023-02-26
Fix unreadable contents during dark mode

JVT038:


JVT038:
Fixed #289.
This is how it looks normally:
image


This is how it looks when hovering an item:

image

PR #298, opened 2023-04-06
Add more relevant information to the app settings page

leepeuker:


JVT038:
Perhaps try to fetch the release notes for the current version and display that on the about page?

leepeuker:
> Perhaps try to fetch the release notes for the current version and display that on the about page?

I think I would like to do it like overseerr, see here:

image

JVT038:
Yeah, that would look good as well!

leepeuker:
Current state here:

image

For adding releases to the settings page I have started https://github.com/leepeuker/movary/pull/299

leepeuker:
There is still a bug in some text colors when changing themes, looking into it

PR #299, opened 2023-04-07
Add latest releases to app settings page

leepeuker:


leepeuker:
Current state:
image

leepeuker:
The backend changed a bit, we only fetch the latest release now and multiple of them to improve app settings page load performance. 

I think it would be better to fetch the latest releases via JS in the frontend when the page is open, so that the request is not impacting backend performance until the page is send to the user. We can display a loading spinner or something like that when fetching the releases to make this more clear to the user.

leepeuker:
Open issues:
- cache the js response in the frontend for a bit (local storage?) to reduce number of requests against github in repeated page reloads (rate limit)
- format the release description (markdown rendering in browser how?)

PR #309, opened 2023-04-11
Add user managment settings page

leepeuker:


leepeuker:
Open issues: When there is a backend error (e.g. user name/email already exist) we report a generic server error in the frontend. It would be better to report a more qualified error message.

JVT038:
It seems that the E-Mail validation only checks if there is an @ present. I can enter the text '@' and it is accepted.

leepeuker:
> ems that the E-Mail validation only checks if there is an @ present. I can enter the text '@' and it is accepted.

true, I did not want to make anything complicated. I could add a check for a letter before/after the @?

JVT038:
I would use `filter_var`. https://www.php.net/manual/en/function.filter-var.php

leepeuker:
I was thinking about frontend validation, but you are right, I am surprised there is seemingly no backend validation at all yet :thinking: 
I guess it would be better to have the same validation in frontend and backend, otherwise we can have the issue that the frontend says a email is valid and the backend rejects it, which is not really user friendly, that would rule out `filter_var`

I think we do not have to do something fancy here, we are not using emails yet anyway :D

JVT038:
I can try and figure out a nice regex string to validate it?

PR #310, opened 2023-04-13
Replace native browser alerts with bootstrap alerts

leepeuker:


leepeuker:
I think I have removed all `alert` method calls except for the one when the movie rating update on a movie page failed, but I was not sure how to best display an alert there. I think I will rebuild the rating edit UI from scratch, I think having a modal pop up to edit the rating is a bit more direct than making the displayed stars editable :thinking:

PR #317, opened 2023-04-15
Update settings account page UX/UI

leepeuker:


JVT038:
Can you add password validation for at least 1 number and 1 special character?

PR #318, opened 2023-04-17
Add watchlist

leepeuker:


JVT038:
Generally looks good, the only things that I noticed was:

1. Shouldn't we use an AJAX / XHR request to add the movie to the watchlist, so the page doesn't refresh? Additionally, there currently isn't any feedback when the movie is added to the watchlist. A success message should be shown if the movie was added or an error message if something went wrong. (Or if the movie was already in the watchlist)
2. When a movie is in the watchlist and I add a 'play', shouldn't the movie be removed from the watchlist automatically? Because I marked the movie as 'to-watch', then I told Movary 'I have watched this movie'. Shouldn't Movary automatically remove the movie from the watchlist then?
3. I feel like we should add a row in the dashboard / homepage for the watchlist, where a limited number of items from the watchlist are shown in the dashboard.

leepeuker:
Thanks for the feedback!

1. I was thinking about this too, but for now I just cloned the way how we are handling logging plays to stay consistent. I think that is fine for now, we can talk about how to handle this better in a new issue?
2. Hm, good question. Maybe we could add a option in the settings to enable/disable automatic removing from the watchlist if a need play was added? Just doing it seems to me like implicit behavior which could irritate the user if he does not know about it :thinking:  I have created an [issue](https://github.com/leepeuker/movary/issues/333)
3. Good idea! I have created an [issue](https://github.com/leepeuker/movary/issues/332)

PR #321, opened 2023-04-21
 Added registration with new environment variable

JVT038:
Opening this PR, because it's not a fork.

Done: 
- [x] Moved create user logic to the
- [x] CreateUserController
- [x] Changed title of create user page to 'create user'
- [x] Added extra route for the create user page

So I have added the posibility to set an environment variable called `REGISTRATION` to 'true' or 'false'. If it's true, the login page will contain a new link below the 'login' button where the user will be redirected to a new create user page. This create user page is rendered from the CreateUserController; it makes a lot more sense to me to move the rendering logic of this page from the LandingPageController to the CreateUserController.


README.md
```diff
@@ -194,6 +194,7 @@ It is recommended to enable tmdb image caching (set env variable `TMDB_ENABLE_IM
 | `LOG_LEVEL`                                 |        `warning`        |                                                                         |
 | `LOG_ENABLE_STACKTRACE`                     |           `0`           |                                                                         |
 | `LOG_ENABLE_FILE_LOGGING`                   |           `1`           | Log directory is at `storage/logs`                                      |
+| `REGISTRATION`                              |          `false`        | Allows anyone to create an account when they're not logged in. Useful for public instances. Allowed values are `true` or `false` |
```
leepeuker, line 197: let us stay with `0`  and `1` like with the other boolean variables to be consistent?

src/HttpController/CreateUserController.php
```diff
@@ -11,20 +11,48 @@
...
         private readonly Authentication $authenticationService,
         private readonly UserApi $userApi,
+        private readonly Config $config,
         private readonly SessionWrapper $sessionWrapper,
     ) {
     }
 
+    public function render() : Response
+    {
+        if ($this->userApi->hasUsers() === true && strtolower($this->config->getAsString('REGISTRATION', 'false')) == 'false' || $this->authenticationService->isUserAuthenticated() === true) {
```
leepeuker, line 31:
```suggestion
        if ($this->userApi->hasUsers() === true && $this->config->getAsBool('REGISTRATION', false)) === false || $this->authenticationService->isUserAuthenticated() === true) {
```

src/HttpController/CreateUserController.php
```diff
@@ -11,20 +11,48 @@
...
+            return Response::createSeeOther('/');
+        }
+
+        $errorPasswordTooShort = $this->sessionWrapper->find('errorPasswordTooShort');
+        $errorPasswordNotEqual = $this->sessionWrapper->find('errorPasswordNotEqual');
+        $errorUsernameInvalidFormat = $this->sessionWrapper->find('errorUsernameInvalidFormat');
+        $errorGeneric = $this->sessionWrapper->find('errorGeneric');
+
+        $this->sessionWrapper->unset('errorPasswordTooShort', 'errorPasswordNotEqual', 'errorUsernameInvalidFormat', 'errorGeneric');
```
leepeuker, line 40: Not to me: I will implement the same I did on the settings pages, so that we do not need these session data anymore

leepeuker: There can be more errors now with a general registration, like the user name already being in use

README.md
leepeuker: hm and I think naming the variable `ENABLE_REGISTRATION` makes it a bit more clear what this means, `REGISTRATION` is a bit to generic :thinking:

src/HttpController/CreateUserController.php
JVT038: I have added the 'missingFormData' error (when there are empty fields), but I haven't found the error for the name being in use.

leepeuker (approved):

src/HttpController/CreateUserController.php
leepeuker: I tried using javascript for validation but it was more effort than I thought, so let us ignore this for now

PR #328, opened 2023-05-01
Add server domain setting

JVT038:
This adds the setting to set a server domain.
Protocol prefix such as http:// or https:// is not allowed.
Only things like my.domain.com is allowed.
The following regex is used to validate it: `/^\w*(\.\w{2,})+/gm`.

Explanation:
`^` = Start of string
`\w` = Any letter, number and underscore (equal to `[a-zA-Z0-9_]`)
`*`= Matches the previous character between zero and unlimited times.
`(` = starts a group
`\.` = matches the dot (for TLDs and subdomains and such)
`\w` = see above
`{2,}` = matches two or more of the previous character. So two or more of `\w`. (TLDs must have at least two characters)
`)` = closes the group
`+` = matches the group at least once. This ensures that things like `asdf` isn't allowed, but `asdf.as` is allowed. Additionally, subdomains are allowed as well as the weird British `.co.uk` thingies are allowed.


leepeuker:
I disagree on not allowing to set the protocol, if a server is using https it should be able set this explicitly, so that no redirects have to be made be the webserver (reduces access log, potential for errors, is more secure).

Additionally I think there are even more potential cases the regex would not allow (e.g. using an IP address if a user does not bother using hostnames). To be honest I think I would not validate this field at all :thinking: 

Examples values I think should be possible:

```
https://sub.domain.tld
http://domain.tld
sub.domain.tld
domain.tld
localhost
192.168.1.5
127.0.0.1
``

Maybe the name `server domain` is wrong for what this config is used for, maybe we should name it `Server Url` instead?

JVT038:
Oh ok, I thought it was just the domain.
But yeah, we should change iti to application url in that case.

Be aware- the regex will become a lot more complicated by adding that.

JVT038:
I think we should ensure the same validation in front- and backend, to keep things consistent.
It's weird to have an error in the frontend that shouldn't exist in the backend or vice versa.
This is the new pattern I came up with: `/^(https?\:\/\/)?[a-zA-Z0-9_-]{0,63}(\.[a-zA-Z0-9_-]0,63})*/`.

It allows for a `http(s)://` prefix, with numbers, letters, underscores and dashes in the (sub)domains.
This also allows IP addresses and things like `localhost`.
BTW, apparently a 'label' has a maximum length of 63 characters. (example: `my.domain.com`, the labels are `my`, `domain` and `com`).
Additionally, dashes and underscores are allowed as well, which is why I replaced the `\w` with the `[a-zA-Z0-9_-]`.
You can test the regex here: https://regex101.com/r/b7vCqy/1

JVT038:
BTW, a danger of my regex and the `filter_var` function of PHP, is that it won't match any non-latin character.
So when you have things like `ü`, `ö` and any special character like that in your domain name, it won't match.

leepeuker:
I would again vote against such a custom regex for validation. What about `sub.sub.domain.tld` and other cases of valid urls we are not thinking of right now which would not work against the current regex?

I see two options here:
- no or very basic validation (e.g. no spaces) which is the same in frontend and backend
- rely on language features (use the `URL` class in [javascript](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) and `filter_var` in php) which are probably not 100% the same but should be similar enough

I prefer the second option I think. I do not see an issue if the the frontend and the backend possible reporting different errors on edge cases.

leepeuker:
@JVT038 I have done a few things:
- change server name to application url
- validate the url in the frontend directly javascript
- display no url and disable all buttons on the webhook setting pages (plex, jellyfin, emby) if no application url is set. show a warning instead that the url is missing
- use the url when to create the webhook urls
- made the webhooks settings pages completely dynamic (no page reload required any more on updates)

Feeback/testing would be appreciated :)

PR #339, opened 2023-05-10
Added icons to remove items from the watchlist

### LINKED ISSUE #334: Add way to remove movie from watchlist directly
Currently you have to go on a movie detail page and press the `More` button to remove a movie from the watchlist. There should be a way to directly on the watchlist page to remove entries.

JVT038:
This PR fixes #334 by adding icons to the top-right of the movie covers to delete items from the watchlist within the watchlist page.
When clicking on the icon a small modal will appear, asking the user for confirmation. If they click 'yes', the movie will be removed from the watchlist, the modal will be closed and a success message will appear below the search field.
Screenshot:

image

Some notes: 
* I've noticed that the watchlist IDs are sometimes called 'movie ID' (in `MovieController.php`) and sometimes 'watchlistEntryId' (in `watchlist.html.twig`). I would prefer more consistency and just call them either `movie ID` in both files or `watchlistEntryId`.
* In the screenshot above, there is a movie with a red background. This makes the trash / remove icon difficult to see; how should we fix this? I don't like a white background, because then there would just be a white square with a trash icon in the center.


leepeuker:
- I do not really like displaying the trash icon all the time on default, that looks weird to me, like some kind of edit mode is active
- I am fine when we use `movieId` instead of `watchlistEntryId`
- I had the same issue when using a different icon, without setting a background I am not sure how to solve this best

I was thinking, maybe we could do it like that: When a user clicks on a movie poster a modal opens, which contains the buttons `Go to movie` (redirects to the movie page) and `Remove from watchlist` (which removes the movie and reloads the page). We could event add a `Add play` button there to so that a user does not have to go anywhere else to that.

leepeuker:
@JVT038 Have a look at my changes: https://github.com/leepeuker/movary/pull/340/files

What do you think of that?

JVT038:
Close this in favour of #334 ?

leepeuker:
I have added a dim effect on the clicked poster, so that we have a visual feedback which poster is related to opend modal

leepeuker (approved):

PR #353, opened 2023-05-18
Add dummy implementation

leepeuker:


leepeuker:
Problems:
- not resizing chart when increasing window size
- bars are too long on bigger screens, need a way to set a max height/length
- label when hovering over a bar should only display the count number

PR #357, opened 2023-05-19
Add dashboard rows customization

JVT038:
This PR enables dashboard settings.
Some things I'd like to discuss:

Currently the row "ID" is literally the thing that the user sees. 
The rowOrder is an array that looks like this: `['Last Plays', 'Most Watched Actors', 'Most Watched Actresses', 'Most Watched Directors', 'Most Watched Genres', 'Most Watched Languages', 'Most Watched Production Companies', 'Most Watched Release Years'];`.

This is because it would otherwise be difficult to show the correct, user-friendly row name. If there were any translations or something, it would've been easier to replace these strings, but now it's just stored like this.

Another thing is naming consistency. The MySQL column that stores the order of the rows is currently called "dashboard_visible_rows" but I don't think that's an appropriate name. Rather, something like "dashboard_order_of_rows" would be better.
But anyway, what should we use? It's currently kind of mixed, where in some cases the row order is called "visibleRows" and in others it's called "rowOrder" (or something like that).

Last but not least, it annoys me that the rows in the dashboard settings aren't completely centered; this is because the buttons on the right side push the text to the left and I haven't found a way to fix this yet. It's currently slightly misaligned.


JVT038 (commented):
Also, the `rowOrder` is currently a constant, but do you think it should be a Dto instead?

leepeuker:
> Rather, something like "dashboard_order_of_rows" would be better.

I agree

> Last but not least, it annoys me that the rows in the dashboard settings aren't completely centered; this is because the buttons on the right side push the text to the left and I haven't found a way to fix this yet. It's currently slightly misaligned.

Yeah I know, that is bothering me too. That happens at a few other places too. Maybe let us create a separate issue for this?

JVT038:
I have centered the rows, but I'm using a 'fake' element to force the centering.
While it doesn't look good in the source code, it does look good in the frontend. 
@leepeuker what do you think?

leepeuker:
@JVT038 Could you move the centering of the rows into a separate branch/PR, than we could merge this without having to wait for customization PR to be done

leepeuker:
So I had time to checkout it. I really like the feature and your first draft. Next up are a few notes from me.

Issues I found:
- after checking out the PR and executing the DB migrations (without changing the settings) the first item on the dashboard looks like it it es extend but it does not show any content. see: 
image
- after updating the settings `Last Plays` is the only one with change title color even it is not active and the font color does not change on theme change: 
image
- the settings can look weird on mobile: 
image
- on the settings page you have to click at the center of the visibility icon for it to work. if you click near the border the button seems to register the click but nothing changes (move the onclick to the button I think)
- on the settings page, let us display the alert about the success of the update at the bottom over the button and not at the top (that is how where we usually display it)

Suggestion:
I think I would prefer using list groups here, imo that stays more consistent with the rest of the design and makes the UI clearer to understand, example: 
image

JVT038:
I fixed the issues you mentioned

JVT038:
@leepeuker Can we merge this?

leepeuker:
> @leepeuker Can we merge this?

I want to test it again before we do that, I am trying to do that soon

src/Service/Dashboard/Dto/DashboardRow.php
```diff
@@ -0,0 +1,113 @@
+<?php declare(strict_types=1);
+
+namespace Movary\Service\Dashboard\Dto;
+
+class DashboardRow
```
leepeuker, line 5: This is class contains the configuration of the dashboard rows centrally at one place. No other place should need to know what id or name a dashboard row has, it is defined here.

leepeuker:
So, I have refactored a few things. Can you have a look at my changes and test them, if you are fine with it I think we can merge it

leepeuker:
Fuck I have clicked the deleted button again accidentally, no idea what is going on today, sorry

src/Service/Dashboard/DashboardFactory.php
```diff
@@ -0,0 +1,61 @@
...
+{
+    public function createDashboardRowsForUser(UserEntity $user) : DashboardRowList
+    {
+        $visibleRows = (string)$user->getDashboardVisibleRows();
+        $extendedRows = (string)$user->getDashboardExtendedRows();
+
+        $visibleRows = $visibleRows !== '' ? explode(';', $visibleRows) : [];
+        $extendedRows = $extendedRows !== '' ? explode(';', $extendedRows) : [];
+
+        if (empty($visibleRows) === true) {
```
JVT038, line 19: Can't you just say `if(empty($visibleRows)) { ... }` without the `===` comparison?

JVT038 (commented):
So I've looked at the changes you've made and I've tested it.

It works, which is the most important thing obviously. I haven't encountered any bugs.
While I can clearly understand what's going on in the code, I feel like it's a bit too complicated. Maybe you're professional and I'm just unprofessional, but I just feel like my version was way less complicated. I don't know if it's a problem; it's not a problem for me, but I think classes like `DashboardRowList` and `DashboardRow` with a bunch of methods and properties are unnecessary when it can just be the same way I did it.

However, despite this, I honestly don't really care lol. I mean, while this may be more complicated, it's also more readable, so that's nice I guess. I honestly just don't care, but the complicated thing was the only 'criticism' that came to my mind while looking through the code.

leepeuker:
> So I've looked at the changes you've made and I've tested it.
> 
> It works, which is the most important thing obviously. I haven't encountered any bugs. While I can clearly understand what's going on in the code, I feel like it's a bit too complicated. Maybe you're professional and I'm just unprofessional, but I just feel like my version was way less complicated. I don't know if it's a problem; it's not a problem for me, but I think classes like `DashboardRowList` and `DashboardRow` with a bunch of methods and properties are unnecessary when it can just be the same way I did it.
> 
> However, despite this, I honestly don't really care lol. I mean, while this may be more complicated, it's also more readable, so that's nice I guess. I honestly just don't care, but the complicated thing was the only 'criticism' that came to my mind while looking through the code.

Yeah maybe my solution is a bit over engineered, but it feels to me like the code is easier extendable in the future and less prone to bugs. Anyway, I was just interested in your opinion, I was not expecting you to care a lot or not :D

src/Service/Dashboard/DashboardFactory.php
leepeuker: Yeah it would not change anything functional, but most of the time it is a difference to use the typesafe `===` operator, so I usually just use it everywhere, so I do not have to think about if a method only returns boolean values

JVT038:
Yeah, your solution is definitely more extendable in the future, that's true.
Anyway, we can merge this right?

leepeuker:
I still have an issue that the `eye` icon and the opacity of the list item display if a row is extended. I think the UI tells the user that this means a row is shown at all, not if it is extended. At least that how I feel when using it.

I suggest that we change the following:
- add a new button/icon to toggle if row is extended
- use the `eye` button and opacity to display if the row is shown in the dashboard at all 

I think we would need a third user option for this. the visible should contain what is visible and we add a new order option :thinking:

leepeuker:
I will create a sub branch with the changes I suggested

leepeuker:
Here is the PR with my suggested changes: https://github.com/leepeuker/movary/pull/390

The new "row extend" button is not perfect yet, only the icon changes to display if is toggled, which is not really that noticeable. 
But the rest works well I think

What do you think about this change?

JVT038:
> the dashboard customization is only for the user himself and is only used on his personal dashboard

I think it should be this one. The user can edit its own dashboard, but not others. Others will view the user's dashboards the way that the user has customized it.

leepeuker:
Alright. So I think that should be everything now :partying_face: 

If you have nothing to not or found another issue, I'd say let's merge the branches together and then into main

leepeuker (dismissed):

leepeuker (dismissed):

leepeuker (approved):

JVT038:
Sure, I got no comments, so we can merge.

PR #367, opened 2023-05-22
Bugfix: Disable unnecessary imagick extension

JVT038:
By adding the environment variable `PHP_DISMOD` and setting it to 'imagick', the container won't install the imagick extension and the error should disappear.
At least, it disappeared on my machine.


leepeuker:
I have moved `PHP_DISMOD=imagick` to the base development image target, so that the development environment has the same extension as the production build to reduce differences

PR #368, opened 2023-05-24
Centered dashboard settings

JVT038:


leepeuker:
@JVT038 I am not sure I get this PR. Imo it should only contain the changes needed to center the dashboard items and it should have the master branch as the target. But the PR contains a lot of other changes. Did I break it somehow when I merged the main branch or am I not seeing something?

JVT038:
So I basically already made a commit to center the dashboard settings, but then I tried to switch that commit to this branch. 
I'm not sure what I did, but I think I kind of broke this branch by trying to move the commit to another branch.

JVT038:
Can't we just delete this branch? Because the settings don't need to be centered anyway, because we've moved to a list view, instead of a navigation.

leepeuker:
> Can't we just delete this branch? Because the settings don't need to be centered anyway, because we've moved to a list view, instead of a navigation.

Fine by me

PR #371, opened 2023-05-29
Add email support basics

leepeuker:


JVT038 (commented):
I think the sending of an email should be scheduled as a job; this way, the user's interface won't be slowed down by having to wait for the email to be sent.

leepeuker:
> I think the sending of an email should be scheduled as a job; this way, the user's interface won't be slowed down by having to wait for the email to be sent.

I think the sending of an email should only take a few seconds max, having to wait so long to get a direct feedback is better than having a job and pressing refresh 500 times. Do you have a different experience or am I mistaken about the duration the sending of an email can take?

JVT038:
I just feel like it would be better if there would be an instant message saying "The E-Mail will be sent soon" instead of waiting a couple of seconds and showing a message "The E-Mail has been sent".

leepeuker:
> I just feel like it would be better if there would be an instant message saying "The E-Mail will be sent soon" instead of waiting a couple of seconds and showing a message "The E-Mail has been sent".

I feel what you mean, but than we need a solution how to show the user the success or failure of it without him having to search for it I think.

But I think we can look into that independently of this scope here. The basic functionally works right now, this would be mainly only an UX change. When we have emails up and running for password resets we can refactor that part, but I would like to focus on that first. We need to add a way to queue email jobs than anyway, which we could reuse then for the test email

PR #386, opened 2023-06-11
Added docker secrets support

JVT038:
I've added Docker secret support for two variables:

* DATABASE_MYSQL_PASSWORD
* DATABASE_MYSQL_ROOT_PASSWORD

It should work in theory, but haven't tested it yet, because I don't like using Docker on Windows 10, and I especially dislike having to debug applications in a docker environment.

### DIFF
--- a/docs/install/docker.md
+++ b/docs/install/docker.md
@@ -118,14 +118,6 @@ volumes:
 ```yaml
 version: "3.5"
 
-secrets:
-  mysql_root_password:
-    file: /path/to/docker/secret/mysql_root_password_file
-  mysql_password:
-    file: /path/to/docker/secret/mysql_password_file
-  tmdb_key:
-    file: /path/to/docker/secret/tmdb_key_file
-
 services:
   movary:
     image: leepeuker/movary:latest
@@ -142,6 +134,7 @@ services:
     volumes:
       - movary-storage:/app/storage
     secrets:
+      - tmdb_key
       - mysql_password
 
   mysql:
@@ -156,6 +149,14 @@ services:
     secrets:
       - mysql_root_password
       - mysql_password
+
+secrets:
+  mysql_root_password:
+    file: /path/to/docker/secret/mysql_root_password
+  mysql_password:
+    file: /path/to/docker/secret/mysql_password
+  tmdb_key:
+    file: /path/to/docker/secret/tmdb_key
 
 volumes:
   movary-db:
--- a/docs/install/docker.md
+++ b/docs/install/docker.md
@@ -31,10 +31,18 @@ The easiest way to do this are managed docker volumes (used in the examples belo
     If you bind a local mount, make sure the directory exists before you start the container
     and that it has the necessary permissions/ownership.
 
+## Docker secrets
+
+Docker secrets can be used for all environment variables, just append `_FILE` to the environment variable name.
+Secrets are used as a fallback for not existing environment variables.
+Make sure to not set the environment variable without the `_FILE` prefix if you want to use a secret.
+
+For more info on Docker secrets, read the [official Docker documentation](https://docs.docker.com/engine/swarm/secrets/).
+
 ## Examples
 
 All examples include the environment variable `TMDB_API_KEY` (get a key [here](https://www.themoviedb.org/settings/api)).
-It is not strictly required to be set here but recommend. 
+It is not strictly required to be set here but recommend.
 Many features of the application will not work correctly without it.
 
 ### With SQLite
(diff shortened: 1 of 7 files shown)


leepeuker:
Hm, in your solution the secrets are overwriting the environment variables, I am not sure if it should maybe be the other way around. The priority could be:
- environment variables (because they are explicitly set for this environment)
- docker secrets
- database (usually config set by user manually via UI/CLI)

If add `-e DATABASE_MYSQL_PASSWORD=foobar` to my container start I think I would expect this explicitly set `foobar` be taken over a implicit secret. What do you think?

JVT038:
Yeah, that's fine by me.

docs/install/docker.md
```diff
@@ -106,3 +106,62 @@ volumes:
   movary-db:
   movary-storage:
 ```
+
+### With Docker secrets
+
+The following environment variable can also be added with secrets.
+
+* `DATABASE_MYSQL_PASSWORD` 
+* `DATABASE_MYSQL_ROOT__PASSWORD`
```
leepeuker, line 115:
```suggestion
* `DATABASE_MYSQL_ROOT_PASSWORD`
```

I think one underscore to much?

src/ValueObject/Config.php
```diff
@@ -14,6 +15,9 @@ public static function createFromEnv(array $additionalData = []) : self
     {
         $fpmEnvironment = $_ENV;
         $systemEnvironment = getenv();
+        // $mysqlRootPassword = file_get_contents($fpmEnvironment['DATABASE_MYSQL_ROOT_PASSWORD_FILE']);
+        // $mysqlPassword = file_get_contents($fpmEnvironment['DATABASE_MYSQL_PASSWORD_FILE']);
+        // array_push($additionalData, [''])
```
leepeuker, line 20:
```suggestion
```

docs/install/docker.md
JVT038: Yup, a small typo, good catch!

settings/phinx.php
```diff
@@ -5,6 +5,12 @@
 
 $config = $container->get(Movary\ValueObject\Config::class);
 
+try {
+    $mysqlPassword = $config->getAsString('DATABASE_MYSQL_PASSWORD');
+} catch(OutOfBoundsException) {
+    $mysqlPassword = $config->getSecretAsString('DATABASE_MYSQL_PASSWORD_FILE');
+}
```
leepeuker, line 12: this logic is duplicated, we should probably refactor this to one place, but I am not sure where to best put it :thinking: maybe let us add a new method to `Config` named `getDatabaseMysqlPassword`, then we could set `getSecretAsString` as private. I think we could encapsulate all config values with getter methods, so we do not need to know how a setting is named in the environment out site of the class. `getAsString` and the rest of the generic methods could be set to private than too

JVT038: Good catch. Yes, I'll move this logic to a separate method in the Config class, but I won't refactor the `getAsString` method in this PR, because then I'll have to change a lot of stuff throughout the application that's not related to Docker secrets, and therefore it's not related to this PR. I (or maybe you or someone else) can do that later, in a separate PR.

leepeuker: I already did it, just pushed it, sorry :D Please have a look, I think I like the solution

src/ValueObject/Config.php
```diff
@@ -57,17 +51,20 @@ public function getAsString(string $parameter, ?string $fallbackValue = null) :
...
+        if (isset($this->config[$key . '_FILE']) === true) {
+            $secretFile = $this->get($key . '_FILE');
+
+            if ($this->fileUtil->fileExists($secretFile) === true) {
+                return $this->fileUtil->readFile($secretFile);
+            }
+        }
 
-        return $this->config[$key];
+        throw new \RuntimeException('Config key does not exist: ' . $key);
```
leepeuker, line 68: We should add a custom exception here to make this more clear I think

settings/phinx.php
JVT038: Ooh, that explains why I just had merge conflicts lol.

JVT038:
I just looked at your changes in the latest commits, and I see nothing wrong with it. I think it's a nice solution :D

leepeuker:
I have updated the docs too. Would be great if you could double check what I wrote :D I have not tested this yet, will try to do this tomorrow. If it works as expected it would be fine for me to merge.

src/ValueObject/Config.php
leepeuker: done

leepeuker:
ups, I accidentally closed the PR, sorry

leepeuker (approved):

PR #393, opened 2023-06-14
Added watchlist to the dashboard rows

JVT038:
Comment: I don't know if the name of the row should be 'Things to watch' or just 'watchlist'.
It's currently 'things to watch'.

Also, really nice API to interact with the watchlist!

### DIFF
--- a/templates/component/dashboard/row-watchlist.html.twig
+++ b/templates/component/dashboard/row-watchlist.html.twig
@@ -1,9 +1,9 @@
-{% macro create(isExtended, watchlistItems, routeUsername) %}
+{% macro create(dashboardRow, watchlistItems, routeUsername) %}
 
     {% import 'component/dashboard/row-item-toggle.html.twig' as rowItemToggle %}
-    {{ rowItemToggle.create('Latest in Watchlist', isExtended) }}
+    {{ rowItemToggle.create(dashboardRow) }}
 
-    <li class="list-group-item {{ isExtended ? 'activeItem' : 'inactiveItem' }}" style="padding: 1rem 0.4rem 0 0.4rem;">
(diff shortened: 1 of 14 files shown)


templates/component/dashboard/row-watchlist.html.twig
```diff
@@ -0,0 +1,27 @@
...
+    {{ rowItemToggle.create('Things to watch', isExtended) }}
+
+    <li class="list-group-item {{ isExtended ? 'activeItem' : 'inactiveItem' }}" style="padding: 1rem 0.4rem 0 0.4rem;">
+        <div class="row row-cols-3 row-cols-md-3 row-cols-lg-6">
+            {% for watchlistItem in watchlistItems %}
+                <div class="col" style="padding-bottom: 1rem;">
+                    <div class="card h-100 position-relative" onclick="window.location='/users/{{ routeUsername }}/movies/{{ watchlistItem.id }}'"
+                         style="cursor: pointer;">
+                        <div class="card-header text-truncate" style="padding: 0.2rem">
+                            <small class="text-muted" style="font-size: 0.9rem;">{{ watchlistItem.title }}</small>
```
leepeuker, line 13: hm, why did you add the title here. do you think we should add the title in the last play row too?

leepeuker:
> Comment: I don't know if the name of the row should be 'Things to watch' or just 'watchlist'. It's currently 'things to watch'.

I have changed it to `Latest in Watchlist`, how do you like that?


> Also, really nice API to interact with the watchlist!

Tanks :pray:

templates/component/dashboard/row-watchlist.html.twig
leepeuker: image

I think having the title makes sense from a usability perspective, having the title in plain text helps to identify a movie in case of missing posters or posters without a title.

still, I think it looks better without it, less texts makes everything look more simpel and clean.

leepeuker: I have removed the title for now to stay consistent. We can still discuss adding it back though.

PR #415, opened 2023-06-28
Refacotring and adjustments

leepeuker:


leepeuker:
I deleted it because it was like you said not in use. I deleted a bunch of code that was not used, I was limiting the scope everywhere to be precise and have less code to read. 
sure we will probably need it later, but than we can just copy and paste it from the history again

PR #422, opened 2023-07-04
Added biography and more details to people

JVT038:
Could you check out the mobile layout? I don't really like it, but don't know how to improve it.


JVT038:
It currently looks like this:

image

The distance / margins between the image and the biography don't change anymore.

leepeuker:
The show more button looks off too I think 

image

And I would prefer if we lose the circle, this does not feel matching to the rest :thinking:

PR #446, opened 2023-07-26
Sort by watched

pbogre:
closes #444

added option in toggleSearchModal "SortBy" field to sort by `watched_at` value in database.

PS: sorry about the mess with the commits

PR #448, opened 2023-07-26
add watched_at to GROUP BY sql entry (fix after #446)

pbogre:
sorry about creating a new pr but i deleted the branch from 446 and cant seem to keep committing to it because it was merged, anyway let me know if this fixes that error for you


pbogre:
just tested and seems to work using sqlite

PR #451, opened 2023-07-27
Save sorting options using cookies

### LINKED ISSUE #441: save sorting options
when i go to the 'all movies' section it automatically sorts them in alphabetical order, and when i set the options so that it sorts by rating in descending order it works fine, however i would like this option to persist even after i leave the page.

maybe this could be done with a 'default sorting' option in the settings, but if that is set to 'sort by rating' it falls back to alphabetical sort for categories that don't have ratings. or save the sorting preference for each page.

pbogre:
use cookies to save sorting options for movies and people.

the cookies are set to last 1 year and are based in "/", here's what an example looks like:

'movie-sort-order' = 'desc'
'movie-sort-by' = 'name'
'person-sort-order' = 'asc'
'person-sort-by' = 'uniqueAppearances'

closes #441

### DIFF
--- a/src/HttpController/ActorsController.php
+++ b/src/HttpController/ActorsController.php
@@ -44,6 +44,9 @@ class ActorsController
         $actorsCount = $this->movieHistoryApi->fetchMostWatchedActorsCount($userId, $requestData->getSearchTerm(), $requestData->getGender());
         $paginationElements = $this->paginationElementsCalculator->createPaginationElements($actorsCount, $requestData->getLimit(), $requestData->getPage());
 
+	setcookie("person-sort-by", $requestData->getSortBy(), time() + 60 * 60 * 24 * 365, "/");
+	setcookie("person-sort-order", (string)$requestData->getSortOrder(), time() + 60 * 60 * 24 * 365, "/");
+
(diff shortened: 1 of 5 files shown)


pbogre:
reminder: create & use save sorting button for people too once a search option modal is implemented

pbogre:
@leepeuker all set with the new modal ui! feel free to take a look using latest commit

leepeuker:
There are a few things I would like to adjust:
- only use cookies for visitors, for logged in users it should be stored in the backend, so that the user does not have to save the option again on every new client
- the options should only be saved if a visitor/user explicitly wants it and not automatically like it is currently (the options should reset on default, at least that is usually the standard in UX imo)
- maybe we should include the filter options as well and not only the sort options 🤔

pbogre:
i agree about using the backend and not cookies for the preferences. maybe a new column in the user settings database?

the options shouldn't be saved automatically in the latest commit. unless you click the button that saves them they shouldn't persist next time you load the page, and if they do that's a bug that i need to fix :) 

i'm not sure about saving other options like filters, because that's not really something that people would want to persist on a long term basis, while sorting options might

PR #453, opened 2023-07-27
remove duplicate movies in "all movies"

### LINKED ISSUE #452: bug - movie entries no longer unique
### Description

i suspect that this is because the SQL query now groups by `watched_at`, which means that there are different entries even when sorting by title, release date, etc.

to fix this, make the sql query sort by the first `watched_at` date.

im not familiar enough with sql to do this right away, but maybe it is possible to remove entries with duplicate title and only leave the one with the earliest watch date?

todo:
- [ ] remove duplicate entries in query, only leave out earliest watch date
- [ ] change `sortBy` name from `Watch date` to `Earliest watch date` 

### Version

latest repo

### Steps to Reproduce

1. build & run latest repo
2. add movie with two different watch dates
3. go to all movies, see 2 entries

### Screenshots

_No response_

### Logs

_No response_

### Server OS

_No response_

### Client Platform

None

### Client Device

_No response_

### Client OS

_No response_

### Client Browser

_No response_

### Additional Context

_No response_

pbogre:
fixes duplicate movies as a result of #446 

closes #452

### DIFF
--- a/src/Domain/Movie/MovieRepository.php
+++ b/src/Domain/Movie/MovieRepository.php
@@ -679,6 +679,7 @@ class MovieRepository
                 LEFT JOIN movie_user_rating mur on mh.movie_id = mur.movie_id and mh.user_id = ?
                 LEFT JOIN movie_genre mg on m.id = mg.movie_id
                 LEFT JOIN genre g on mg.genre_id = g.id
+                $whereQuery
                 GROUP BY m.id, title, release_date, watched_at, rating
                 ORDER BY $sortBySanitized $sortOrder, title asc
                 LIMIT $offset, $limit
--- a/src/Domain/Movie/MovieRepository.php
+++ b/src/Domain/Movie/MovieRepository.php
@@ -672,7 +672,7 @@ class MovieRepository
 
         return $this->dbConnection->fetchAllAssociative(
             <<<SQL
-            SELECT m.*, mur.rating as userRating
+            SELECT DISTINCT m.*, mur.rating as userRating
             FROM movie m
             JOIN movie_user_watch_dates mh on mh.movie_id = m.id and mh.user_id = ?
             LEFT JOIN movie_user_rating mur ON mh.movie_id = mur.movie_id and mur.user_id = ?
--- a/src/Domain/Movie/MovieRepository.php
+++ b/src/Domain/Movie/MovieRepository.php
@@ -672,16 +672,18 @@ class MovieRepository
 
         return $this->dbConnection->fetchAllAssociative(
             <<<SQL
-            SELECT DISTINCT m.*, mur.rating as userRating
-            FROM movie m
-            JOIN movie_user_watch_dates mh on mh.movie_id = m.id and mh.user_id = ?
-            LEFT JOIN movie_user_rating mur ON mh.movie_id = mur.movie_id and mur.user_id = ?
-            LEFT JOIN movie_genre mg on m.id = mg.movie_id
-            LEFT JOIN genre g on mg.genre_id = g.id
-            $whereQuery
-            GROUP BY m.id, title, release_date, watched_at, rating
-            ORDER BY $sortBySanitized $sortOrder, title asc
-            LIMIT $offset, $limit
+            SELECT * FROM (
+                SELECT m.*, mur.rating as userRating, ROW_NUMBER() OVER(PARTITION BY m.id) rn
+                FROM movie m
+                JOIN movie_user_watch_dates mh on mh.movie_id = m.id and mh.user_id = ?
+                LEFT JOIN movie_user_rating mur on mh.movie_id = mur.movie_id and mh.user_id = ?
+                LEFT JOIN movie_genre mg on m.id = mg.movie_id
+                LEFT JOIN genre g on mg.genre_id = g.id
+                GROUP BY m.id, title, release_date, watched_at, rating
+                ORDER BY $sortBySanitized $sortOrder, title asc
+                LIMIT $offset, $limit
+            ) a
+            WHERE rn = 1
             SQL,
             $payload,
         );
--- a/src/Domain/Movie/MovieRepository.php
+++ b/src/Domain/Movie/MovieRepository.php
@@ -682,9 +682,9 @@ class MovieRepository
                 $whereQuery
                 GROUP BY m.id, title, release_date, watched_at, rating
                 ORDER BY $sortBySanitized $sortOrder, title asc
-                LIMIT $offset, $limit
             ) a
             WHERE rn = 1
+            LIMIT $offset, $limit
             SQL,
             $payload,
         );
--- a/templates/page/movies.html.twig
+++ b/templates/page/movies.html.twig
@@ -32,7 +32,7 @@
                         <select class="form-control" name="sb" id="sort-by" style="width: 50%">
                             <option value="title" {{ sortBy == 'title' ? 'selected' }}>Title</option>
                             <option value="releaseDate" {{ sortBy == 'releaseDate' ? 'selected' }}>Release date</option>
-                            <option value="watchDate" {{ sortBy == 'watchDate' ? 'selected' }}>Watch date</option>
+                            <option value="watchDate" {{ sortBy == 'watchDate' ? 'selected' }}>First watch date</option>
                             <option value="rating" {{ sortBy == 'rating' ? 'selected' }}>Rating</option>
                             <option value="runtime" {{ sortBy == 'runtime' ? 'selected' }}>Runtime</option>
                         </select>


src/Domain/Movie/MovieRepository.php
```diff
@@ -672,7 +672,7 @@ public function fetchUniqueMoviesPaginated(
 
         return $this->dbConnection->fetchAllAssociative(
             <<<SQL
-            SELECT m.*, mur.rating as userRating
+            SELECT DISTINCT m.*, mur.rating as userRating
```
leepeuker, line 675:
```suggestion
            SELECT DISTINCT m.*, mur.rating as userRating, watched_at
```

this fixes it

templates/page/movies.html.twig
```diff
@@ -32,7 +32,7 @@
                         <select class="form-control" name="sb" id="sort-by" style="width: 50%">
                             <option value="title" {{ sortBy == 'title' ? 'selected' }}>Title</option>
                             <option value="releaseDate" {{ sortBy == 'releaseDate' ? 'selected' }}>Release date</option>
-                            <option value="watchDate" {{ sortBy == 'watchDate' ? 'selected' }}>Watch date</option>
+                            <option value="watchDate" {{ sortBy == 'watchDate' ? 'selected' }}>First watch date</option>
```
leepeuker, line 35:
```suggestion
                            <option value="watchDate" {{ sortBy == 'watchDate' ? 'selected' }}>Watch date</option>
```

I do no think that `First` improves the meaning or is maybe even correct :thinking: Which watch date is relevant, if there are multiple for a movie, the user decides by the sort order selection. So relevant could be the first watch date or the last, it differs for use cases. Labeling the option just `Watch date` seems better fitting to me here, because it is more generic

src/Domain/Movie/MovieRepository.php
pbogre: the problem with this fix is that it selects entries with distinct `watched_at` values, which means that there will still be entries for each watch date of a movie (at least with sqlite)

templates/page/movies.html.twig
leepeuker: Hm, I think I see what you mean, but I still believe that these prefixes do not provide really necessary information. Sorting by `Watch date` in `Ascending ` order implies that the oldest date is display first and sorting by `Descending` the latest first without the need for more description. Additionally naming it only `Watch date` stays consistent with the rest of the sort options like `Release date` or `Title`, which are simply named after the attribute they represent.

pbogre: yeah thats fair enough

src/Domain/Movie/MovieRepository.php
pbogre: ok so i got an SQL query that seems promising and i tested with both mysql and sqlite outside of movary (using a movary database), however there seems to be an error when i use it with movary in `MovieRepository.php` which you can see below

```
An exception occurred while executing a query: Unable to bind parameter number 3 (25) at /app/vendor/doctrine/dbal/src/Driver/API/SQLite/ExceptionConverter.php:83)
```

here is the sql query that seems to work when ran outside movary using a movary db:

```sql
SELECT * FROM (
     SELECT m.*, mur.rating as userRating, ROW_NUMBER() OVER(PARTITION BY m.id) rn
     FROM movie m
     JOIN movie_user_watch_dates mh on mh.movie_id = m.id
     LEFT JOIN movie_user_rating mur ON mh.movie_id = mur.movie_id
     LEFT JOIN movie_genre mg on m.id = mg.movie_id
     LEFT JOIN genre g on mg.genre_id = g.id
) a
WHERE rn = 1
ORDER BY title asc;
```

It is possible that the way i implemented it in `MovieRepository.php` is incorrect (seen below), since i'm not 100% on how sql query processed and handled here. if you know of any way that this could be implemented correctly that would be awesome.

Screenshot 2023-07-27 at 19 29 00

pbogre: i did some more digging and this version seems to solve the problem from earlier

```sql
<<<SQL
SELECT * FROM (
     SELECT m.*, mur.rating as userRating, ROW_NUMBER() OVER(PARTITION BY m.id) rn
     FROM movie m
     JOIN movie_user_watch_dates mh on mh.movie_id = m.id
     LEFT JOIN movie_user_rating mur ON mh.movie_id = mur.movie_id and mur.user_id = ?
     LEFT JOIN movie_genre mg on m.id = mg.movie_id
     LEFT JOIN genre g on mg.genre_id = g.id
     GROUP BY m.id, title, release_date, watched_at, rating
     ORDER BY $sortBySanitized $sortOrder, title asc
     $whereQuery
     LIMIT $offset $limit
) a
WHERE rn = 1
SQL,
$payload,
```

now i'm getting another error which is more readable thankfully: 
```
An exception occurred while executing a query: Unable to prepare statement: 1, near "WHERE": syntax error {"exception":"[object] (Doctrine\\DBAL\\Exception\\SyntaxErrorException
```

pbogre: new query version that includes GROUP BY and corrects ORDER BY location

```sql
SELECT * FROM (
   SELECT m.*, mur.rating as userRating, ROW_NUMBER() OVER(PARTITION BY m.id) rn
   FROM movie m
   JOIN movie_user_watch_dates mh on mh.movie_id = m.id
   LEFT JOIN movie_user_rating mur ON mh.movie_id = mur.movie_id
   LEFT JOIN movie_genre mg on m.id = mg.movie_id
   LEFT JOIN genre g on mg.genre_id = g.id
   GROUP BY m.id, title, release_date, watched_at, rating
   ORDER BY watched_at desc, title asc
   ) a
WHERE rn = 1;
```

tested with sqlite and mysql outside movary, still getting that parameter error when running inside movary

leepeuker:
```            JOIN movie_user_watch_dates mh on mh.movie_id = m.id and mh.user_id = ?
            LEFT JOIN movie_user_rating mur ON mh.movie_id = mur.movie_id and mur.user_id = ?
```

you are missing the `user_id = ?`, maybe that is the issue? ideally push your current state, than I can have a better look

pbogre: sorry about that i was just omitting the user id check for local sql tests. i'll push my latest commit so you can have a look

pbogre: ok so you were actually right about that but i also forgot that i had removed the `$whereQuery`, so after putting both the user_id and whereQuery it seems to be working fine! please check if it also works with mysql on your setup

src/Domain/Movie/MovieRepository.php
```diff
@@ -672,16 +672,19 @@ public function fetchUniqueMoviesPaginated(
...
+                JOIN movie_user_watch_dates mh on mh.movie_id = m.id and mh.user_id = ?
+                LEFT JOIN movie_user_rating mur on mh.movie_id = mur.movie_id and mh.user_id = ?
+                LEFT JOIN movie_genre mg on m.id = mg.movie_id
+                LEFT JOIN genre g on mg.genre_id = g.id
+                $whereQuery
+                GROUP BY m.id, title, release_date, watched_at, rating
+                ORDER BY $sortBySanitized $sortOrder, title asc
+                LIMIT $offset, $limit
+            ) a
+            WHERE rn = 1
```
leepeuker, line 687:
```suggestion
            ) a
            WHERE rn = 1
            LIMIT $offset, $limit
```

We nearly got it I think :D We need to move the offset and limit out of the subquery, otherwise the subquery returns e.g. the wanted 24 results and than removes the results not matching the where -> we have less results than wanted which causes something like this:
image

leepeuker: hm but this still does not seem right :thinking: 

maybe we just solve this via code instead of query, not as elegant but simpler I guess :D

leepeuker: Ah shit via code does not work either because when we get the max amount of results we cannot just remove something from it :hankey:

pbogre: hm, im not getting that bug using sqlite.

the limit line is inside the subquery because that's where all the data is fetched, and i think placing it outside of that would query all of the movies for each page and THEN filter them.

pbogre: which movies are being ignored by the query? i feel like this has to do with the the whole query in general rather than the LIMIT line, unfortunately.

since i'm not running mysql it's probably easier for you to debug which entries are being ignored, also because i don't have access to a computer now.

pbogre: oh ok i see what you mean now, and i was able to replicate the bug with sqlite as well.

you're right about the LIMIT line being the culprit here, it seems that we would have to increment the offset/limit by 1 for each duplicate entry.


i tried applying the fix that you suggested and... it seems to work for me. all 24 slots are taken even with movies that have various watch dates.

leepeuker: Maybe let's do it like this: We have to queries, one with watch date sorting (where we have to fetch all movies every time) and one without. We only use the watch date sorting query when necessary. This way we can lessen the impact of this performance issue.

pbogre: to be fair this change would complicate the way sort-by is handled now, and the performance issue would still be there for everyone who sorts by watch date.

i feel like it is a generally negligible performance issue for the scope of this query, because you would need a LOT of entries for it to slow down the response time. we could of course run some tests with this but i thinks we should stick with the current way that sort-by is handled to avoid the added complexity just for one more way to sort movies.

pbogre: just to make sure, could you try using the latest commit and see if you still get ghost entries

leepeuker: The ghost entries seem gone to me :heavy_check_mark: 

I just checked the query, on my newish pc the original query takes ~0.016 and the new one 0.025 (user had 860 watch dates). I am not sure how well this scales with increased traffic and/or bigger libraries, especially on more low level hardware.
True, probably nothing to worry about most of the time, but still sucks :D

pbogre: nice to see its working.

yeah the performance decrease is pretty much negligible, especially considering that 860 watch dates is a lot for most people. for now it can be something to keep in mind if problems about performance arise in the future

leepeuker: I saw people with 1-2k entries when debugging trakt issues, some even more, you should not underestimate how people use this. but this are of course still low numbers.

PR #456, opened 2023-07-28
Add Jellyfin authentication

JVT038:
This PR adds authentication for Jellyfin, which can be used in future PRs to use and manipulate Jellyfin data.


JVT038:
To check whether the Jellyfin server actually exists, `/system/info/public` is used as an endpoint. This endpoint is accessible without authentication and just returns some basic info about the server (such as the server version and the server name).

To get an access token, [AuthenticateByName](https://api.jellyfin.org/#tag/User/operation/AuthenticateUserByName) will be used. (This is the same endpoint which the Jellyfin web interface uses to authenticate).
The main disadvantage of this endpoint is that it sends the data unencrypted (if the user doesn't use SSL for their Jellyfin server). I would have preferred to use some kind of OAuth2 system (like we do with Plex), but Jellyfin doesn't support that (as far as I know).

JVT038 (commented):
I have added the ability to authenticate with Jellyfin, remove the authentication (which also deletes the access token on the Jellyfin server!) and verify if the Jellyfin server URL is valid.

JVT038:
BTW, I forgot to mention that this PR adds a new environment variable.

`JELLYFIN_DEVICEID` is a randomly generated string of any length (I use 32 characters) and this is how Jellyfin distinguishes different apps using a Jellyfin account in the 'devices' tab of Jellyfin. 

You need to set this environment variable BEFORE authenticating; maybe we should add a verification, to check if this environment variable is set, before trying to authenticate with Jellyfin.

docs/configuration.md
```diff
@@ -20,14 +20,13 @@ The `Web UI` column is set to yes if an environment variable can alternatively b
...
-| `PLEX_IDENTIFIER`                           |        -        | Required for Plex Authentication. Generate with e.g. `openssl rand -base64 32` |        |
-| `PLEX_APP_NAME`                             |    `Movary`     | Required for Plex Authentication                                               |        |
 | `ENABLE_REGISTRATION`                       |       `0`       | Enables public user registration                                               |        |
 | `MIN_RUNTIME_IN_SECONDS_FOR_JOB_PROCESSING` |      `15`       | Minimum time between background jobs processing                                |        |
 | `TIMEZONE`                                  | `Europe/Berlin` | Supported timezones [here](https://www.php.net/manual/en/timezones.php)        |        |
 | `DEFAULT_LOGIN_EMAIL`                       |        -        | Email address to always autofill on login page                                 |        |
 | `DEFAULT_LOGIN_PASSWORD`                    |        -        | Password to always autofill on login page                                      |        |
 
+
```
leepeuker, line 29:
```suggestion
```

PR #458, opened 2023-07-29
Add jellyfin authentication

JVT038:


public/js/settings-integration-plex.js
JVT038, line 1: What is this?

leepeuker: I reverted this, was wrong

leepeuker:
Notes:
- improve connection test message, differentiate between server url and auth success/errors
- improve error messages in general

leepeuker:
I was thinking how we could improve the `Test connection` feature. Maybe we could test for two separate things:
- that the server url exists and can be connected to
- that the authentication is correct

Resulting alerts of the test:
- server url does not exist: display error ("Server exists but authentication not set or invalid")
- server url does exist but auth does not wotk/exist: display warning ("Server exists but authentication not set or invalid")
- server url does exist and auth does work: success

this way the user can get independent results for the server url and the auth :thinking:

leepeuker (approved):

PR #460, opened 2023-07-29
use text for ascending/descending option in search options modal

pbogre:
for consistency with the other options and to use more of the width that is available.

- [x] movie modal
- [x] people modal


leepeuker:
Added people modal here https://github.com/leepeuker/movary/pull/461

leepeuker (approved):

PR #463, opened 2023-07-30
Fix duplication issue when updating existing watchdate with new date

### LINKED ISSUE #438: Updating watch date issue
When updating an existing watch date with a new date, the existing date is correctly duplicated with the new date, but the original date is not deleted.

Android, Chrome PWA

JVT038:
This PR should fix #438.

The issue was that updating the watchdate basically creates a new watch entry. The old watch entry (with the former watch date) wouldn't be removed.
This PR changes this, and it removes the old watch entry, therefore fixing the issue.

### DIFF
--- a/public/js/movie.js
+++ b/public/js/movie.js
@@ -102,7 +102,8 @@ function editWatchDate() {
         url: apiUrl,
         type: 'POST',
         data: JSON.stringify({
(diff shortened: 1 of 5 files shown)


leepeuker (approved):
nice :+1:

PR #468, opened 2023-08-01
Add jellyfin watch state sync

leepeuker:


leepeuker:
- [ ] add cli command to flush/refresh jellyfin cache

PR #470, opened 2023-08-05
Add jellyfin import

leepeuker:


JVT038 (commented):
Well, that was faster than expected.

One thing I'm not sure about is whether we also want to fetch the IMDB ID from Jellyfin. It's currently only the TMDB ID, but we can obviously easily add the IMDB ID as well.

Testing would be appreciated :D

PR #483, opened 2023-08-28
Add basic admin password reset

leepeuker:


JVT038:
Idk if you have already implemented this, but I think it would be a good idea to allow the admin to change their password through the CLI as well.

leepeuker:
> Idk if you have already implemented this, but I think it would be a good idea to allow the admin to change their password through the CLI as well.

Is already part of the CI :+1:

PR #484, opened 2023-08-31
Added a basic table view with cookies at watchlist

JVT038:
This is currently marked as draft, because I don't know which columns should be present.

There are currently only two columns: the movie title and the release year. Which other columns should be shown? I've thought about adding the movie poster / thumbnail.


jvgoch:
Personally I would really like a runtime column. I'm looking for a self hosted watch list app where I can just sort my watchlist by runtime and get a good overview. So far Movary comes the closest.

PR #486, opened 2023-09-02
Add rest api basics

leepeuker:
You can try out the new swagger UI for openapi schema at for example http://127.0.0.1/docs/api

You can get the openapi schema as json from http://127.0.0.1/api/openapi.json


leepeuker:
Done: 
- setup new `/api` namespace
- added first api endpoint `/api/users/{username}/history`
- started with openai schema for the first endpoint
- added swagger ui

templates/page/api.html.twig
```diff
@@ -0,0 +1,21 @@
...
+<html lang="en">
+<head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <title>Movary API</title>
+    <link rel="stylesheet" href="/css/swagger-ui-4.5.0.css" />
+</head>
+<body>
+<div id="swagger-ui"></div>
+<script src="/js/swagger-ui-bundle-4.5.0.js" crossorigin></script>
```
JVT038, line 11: As I mentioned before, it would be better to install swagger as an NPM package; maybe this is a good opportunity to do so?

settings/routes.php
```diff
@@ -1,550 +1,146 @@
...
-        'DELETE',
-        '/api/users/{userId:\d+}',
-        [\Movary\HttpController\Web\UserController::class, 'deleteUser'],
-    );
-};
+function addApiRoutes(FastRoute\RouteCollector $routeCollector) : void
+{
+    $routeCollector->addRoute('GET', '/openapi.json', [Api\OpenApiController::class, 'getSchema']);
+    $routeCollector->addRoute('GET', '/users/{username:[a-zA-Z0-9]+}/history', [Api\HistoryController::class, 'getHistory']);
+}
```
JVT038, line 146: If you're going to add authentication, try to do that by adding middleware in the routes here, instead of adding an if statement in every controller method in the ApiController.

templates/page/api.html.twig
leepeuker: sadly no, I am glad I have time to implement this at all, the additional hours it would probably take to set npm up and configure everything is not in my budget

settings/routes.php
leepeuker: I have not time to look into how to integrate a middleware here atm

leepeuker:
So I am done with the main implementation as far as I see. I will refactor the `HistoryController`, especially the user authentication stuff, before merging this.

PR #490, opened 2023-09-03
Added middleware

JVT038:
While the name of this branch is 'add-web-authentication-middleware', this PR also adds a bunch of other middleware outside of authentication (such as checking whether the user has a Jellyfin / Plex token or is an Admin, etc.)

The middleware stuff is in `src/HttpController/Middleware`.

The new syntax for a route is as follows:
`$routeCollector->addRoute('GET', '<URI>', [[<Class\SomeHandler::class>, '<someMethod>'], 'middleware' => [<someMiddleware::class>,<someOtherMiddleware::class>]]);`.

I'll try and see if I can improve the syntax, because it's currently quite ugly IMO. If no middleware is necessary, the old syntax can be used fine; so then the handler shouldn't be in an array.


JVT038:
Also, I made some new files in the Service directory; that's my attempt at extending the Fastroute library and trying to improve the syntax with some `addMiddleware` method. It can be ignored for now, because it doesn't work at all (and it's not used either).

JVT038:
So I managed to improve the syntax a bit. This is how it currently looks like:

```php
$routes->createNewRoute('GET', '/', [Some\Controller::class, 'callbackfunction'])->addMiddleware(Middleware\SomeMiddleware::class, Middleware\SomeOtherMiddleware::class);
```

JVT038:
I'm not sure if we should replace this check with a middleware:

```php
        $userId = $this->userPageAuthorizationChecker->findUserIdIfCurrentVisitorIsAllowedToSeeUser((string)$request->getRouteParameters()['username']);
        if ($userId === null) {
            return Response::createNotFound();
        }
```
The `$userId` variable is used often and if we were going to replace this with a middleware, the controller would still need the userId and the check method `findUserIdIfCurrentVisitorIsAllowedToSeeUser()` won't be removed. Wouldn't it be unnecessary to replace this thing?

PR #492, opened 2023-09-05
Stop using application url for openapi.json

leepeuker:


JVT038:
image

Doesn't fix anything.

However, if I move the openapi.json file to `/public/api/openapi.json`, it does work.

leepeuker:
Hm I see. I am not sure if I want to place this in the public directory, this feels wrong somehow, I much prefer having this in the docs directory. Maybe we can symlink it :thinking:

PR #496, opened 2023-09-07
Add some documentation for interested developers

JVT038:


JVT038:
@leepeuker There are two things that still need proper explanation / documentation before I can mark this as 'ready for review' and merge it.

Could you add the explanation about for the `/src/Domains` directory (just say what the directory is and what it's meant for) and explain how the dependency injection / bootstrapping works?

leepeuker:
I have adjusted the text at a few places and added an explanation for the `/src/Domains` directory

leepeuker:
@JVT038 I would be fine with it like this for a first version. If you agree and my additions are understandable you can merge this

leepeuker (approved):

PR #500, opened 2023-09-09
Add a Radarr feed URL for watchlist sync

JVT038:
The radarr feed URL works similar to the webhook URLs.

The URL looks like `localhost/api/radarr/<UUID>` and the UUID is stored in the users table.

The URL responds with a JSON list of all the watchlist items. It looks like this:

```json
[
    {
        "title": "ThanksKilling",
        "tmdb_id": 60366,
        "imdb_id": "tt1129441",
        "added_at": "2023-05-20 20:37:30"
    },
]
```

The user can add this list quite easily by going to Radarr, add it as a custom StevenLu list and enter the feed URL from Movary.


settings/routes.php
```diff
@@ -208,6 +210,9 @@ function addApiRoutes(RouterService $routerService, FastRoute\RouteCollector $ro
     $routes->add('DELETE', $routeUserPlayed, [Api\PlayedController::class, 'deleteFromPlayed'], [Api\Middleware\IsAuthorizedToWriteUserData::class]);
     $routes->add('PUT', $routeUserPlayed, [Api\PlayedController::class, 'updatePlayed'], [Api\Middleware\IsAuthorizedToWriteUserData::class]);
 
+    $routes->add('GET', '/radarr/{id:.+}', [RadarrController::class, 'renderRadarrFeed']);
```
leepeuker, line 213: Similar to [this](https://github.com/leepeuker/movary/pull/510/files#r1341935732) topic I am thinking about what the past path would be. Maybe we could group all feeds under the same route `/api/feed/radarr/{id:.+}`

JVT038: I agree, that's a better path.

leepeuker (approved):
Looks good to me now. You can merge if you agree

leepeuker:
We should add the new radarr integration to the docs

leepeuker (approved):

PR #510, opened 2023-09-27
Move webhook endpoints to the REST API

JVT038:
This PR moves the webhooks to the API group instead of web.

Should the other webhook-related functions (such as deleting a webhook URL and regenerating an URL) be moved to the API group as well?


JVT038:
Should the webhook payload also be added to the openapi specs?

leepeuker:
> Should the webhook payload also be added to the openapi specs?

Hm, good question. I guess we could document what we really use from the payload (instead of everything the webhhook source provides). On the other site, documenting it would make it easier for people to use the webhook endpoints for other stuff and I do not think I would like to support this. The webhooks are depended on what the third party sends to us and not on how we would like it.

settings/routes.php
```diff
@@ -210,5 +203,12 @@ function addApiRoutes(RouterService $routerService, FastRoute\RouteCollector $ro
 
     $routes->add('GET', '/movies/search', [Api\MovieSearchController::class, 'search'], [Api\Middleware\IsAuthenticated::class]);
 
+    #####################
+    # Webhook listeners #
+    #####################
+    $routes->add('POST', '/plex/{id:.+}', [Api\PlexController::class, 'handlePlexWebhook']);
+    $routes->add('POST', '/jellyfin/{id:.+}', [Api\JellyfinController::class, 'handleJellyfinWebhook']);
+    $routes->add('POST', '/emby/{id:.+}', [Api\EmbyController::class, 'handleEmbyWebhook']);
```
leepeuker, line 211: Hm, I am not sure how I would prefer the route to look like. I see these options:

- `/api/plex/{id:.+}` -> not really clear what this is about
- `/api/plex/webhook/{id:.+}` -> makes it easier to extend the `plex` namespace with more endpoints which are not related to webhooks
- `/api/webhook/plex/{id:.+}` -> if we have other plex stuff not related to the webhook we would still need a `/api/plex` namespace for it and we would have split the plex specific stuff into two different base routes

I think I prefer the last option, but I cannot really argue why :thinking:

settings/routes.php
```diff
@@ -33,13 +33,6 @@ function addWebRoutes(RouterService $routerService, FastRoute\RouteCollector $ro
     ]);
     $routes->add('GET', '/docs/api', [Web\OpenApiController::class, 'renderPage']);
 
-    #####################
-    # Webhook listeners #
-    #####################
-    $routes->add('POST', '/plex/{id:.+}', [Web\PlexController::class, 'handlePlexWebhook']);
-    $routes->add('POST', '/jellyfin/{id:.+}', [Web\JellyfinController::class, 'handleJellyfinWebhook']);
-    $routes->add('POST', '/emby/{id:.+}', [Web\EmbyController::class, 'handleEmbyWebhook']);
```
leepeuker, line 41: we could maybe deprecate the old endpoints instead of removing them (+ log a warning if they are still used) and wait a few releases until removing them completely. this would probably reduce the risk of breaking the setup for people. Sure we are not stable yet, but would probably still be nice I guess

settings/routes.php
JVT038: I actually thought about this too and I think `/api/webhook/plex/{id:.+}` is the best option, simply because it looks fancy :)

Also, it allows us to add other webhooks for  apps outside of Plex (such as the Jellyfin or Emby webhook) to it by just doing `/api/webhook/{other app}/{id:.+}`.

settings/routes.php
JVT038: How would you warn the users in Movary? 

I don't think it's bad, just say in the release notes that there have been some breaking changes.

leepeuker: Warning simply via the logs. Only is useful if the user is monitoring the logs I guess, but better than nothing

JVT038: Sure, alright I guess. I honestly don't really care lol

leepeuker (approved):
@JVT038 I have pushed my changes, looks good to merge to me now

settings/routes.php
leepeuker: great, `webhook` it is

PR #517, opened 2023-09-30
Improve the handling of unauthenticated users

JVT038:
This PR automatically redirect user to the login page if they're not authenticated with a warning that they have to login before viewing the page.

Besides that, the user will be redirected back to the page they attempted to view if they log in.

### DIFF
--- a/src/HttpController/Web/AuthenticationController.php
+++ b/src/HttpController/Web/AuthenticationController.php
@@ -37,11 +37,12 @@ class AuthenticationController
         } catch (InvalidCredentials) {
             $this->sessionWrapper->set('failedLogin', true);
         }
-
+        $redirect = $postParameters['redirect'];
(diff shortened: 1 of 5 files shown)


JVT038:
BTW, currently the redirection target URL encoded and stored in the GET parameters.
So it's like `localhost/login?redirect={target url}`.

In theory, anyone could just manually edit the target URL in the get parameter and be redirected to whatever they want. Is this a danger, or should we check if the target url exists and stuff?

leepeuker:
> BTW, currently the redirection target URL encoded and stored in the GET parameters. So it's like `localhost/login?redirect={target url}`.
> 
> In theory, anyone could just manually edit the target URL in the get parameter and be redirected to whatever they want. Is this dangerous, or should we check if the target url exists and stuff?

we should make sure to only allow relative urls, I think than we are fine

PR #540, opened 2023-10-28
Change the poster by using TMDB search results

JVT038:
This PR adds the functionality to change the poster by fetching new posters from TMDB and allowing the user to select the desired poster.


templates/page/movie.html.twig
```diff
@@ -107,6 +107,9 @@
                             class="fw-lighter">/{{ movie.tmdbRatingVoteCount|default('-') }}</span>
                 </p>
             {% endif %}
+            {% if canChangePoster == true %}
+                <button class="btn btn-outline-secondary" data-bs-toggle="tooltip" data-bs-title="Change movie poster" onclick="toggleChangeMoviePosterModal()"><i class="bi bi-pencil-square"></i></button>
+            {% endif %}
```
leepeuker, line 112: please remove this button, I do not think we want more buttons on the movie pages
instead, edit the `more-options` modal and add a new button `Update poster` to it, there you can than call your new modal

src/Api/Tmdb/TmdbApi.php
```diff
@@ -59,6 +59,11 @@ public function getWatchProviders(int $tmdbId, string $country) : TmdbWatchProvi
         );
     }
 
+    public function getImages(int $tmdbId) : array
+    {
+        return $this->client->get("/movie/$tmdbId/images");
```
leepeuker, line 64: can we filter the posters somehow? a lot of results are in languages I am not interested in, which makes looking through it pretty cumbersome.

templates/component/modal-change-movie-poster.html.twig
```diff
@@ -0,0 +1,35 @@
...
+        <div class="modal-content">
+            <div class="modal-header">
+                <h1 class="modal-title fs-5" id="changePosterModalHeader"></h1>
+                <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
+            </div>
+            <div class="modal-body">
+                <div class="d-flex flex-sm-row flex-xs-column justify-content-around" id="PosterOptionsContainer">
+                    <div class="PosterOptionsDiv" onclick="toggleChangePosterSection(this)" id="searchTMDBOption">Search TMDB for posters <br/> <i class="bi bi-search"></i></div>
+                    <div class="PosterOptionsDiv" onclick="toggleChangePosterSection(this)" id="pasteUrlOption">Manually paste an URL to a poster <br/> <span style="font-style: italic; ">https://</span></div>
+                    <div class="PosterOptionsDiv" onclick="toggleChangePosterSection(this)" id="uploadImageOption">Manually upload your own poster <br/> <i class="bi bi-upload"></i></div>
```
leepeuker, line 13: please hide/disable these buttons, they should not be part of this branch, this can be moved to a followup branch

public/js/movie.js
```diff
@@ -340,6 +340,123 @@ function refreshWhereToWatchModal() {
...
+    document.getElementById('PosterOptionsContainer').classList.remove('d-none');
+    document.getElementById('searchTMDBSection').classList.add('d-none');
+    document.getElementById('pasteUrlSection').classList.add('d-none');
+    document.getElementById('uploadImageSection').classList.add('d-none');
+    document.getElementById('returnToOptionsBtn').classList.add('d-none');
+}
+
+function processTMDBPosters(TMDBData) {
+    document.getElementById('TMDBPosterResults').innerHTML = '';
+    TMDBData.forEach(imageData => {
```
leepeuker, line 377: we should not load all images, doing this for a few movies can cause rate limit issues quickly I think.
we should load only first X (10?) images and more only when necessary (user scrolls down or presses a button "show more")

src/Domain/Movie/MovieApi.php
```diff
@@ -532,6 +535,19 @@ public function updateLetterboxdId(int $movieId, string $letterboxdId) : void
...
 
+    public function updatePosterPath(int $movieId, string $posterPath, ?string $oldPosterPath) : void
+    {
+        if($oldPosterPath !== null) {
+            $this->imageCacheService->deleteImageByPosterPath($oldPosterPath);
+        }
+        $this->repository->updateTmdbPosterPath($movieId, $posterPath);
+
+        $tmdbPosterUrl = $this->tmdbUrlGenerator->generateImageUrl($posterPath);
+        $newPosterPath = $this->imageCacheService->cacheImage($tmdbPosterUrl, $movieId, ResourceType::createMovie(), true);
```
leepeuker, line 546: what if the user does has image caching disabled? I think you are missing this check here

src/Domain/Movie/MovieRepository.php
```diff
@@ -1054,6 +1054,20 @@ public function updateLetterboxdId(int $id, string $letterboxdId) : void
...
+    {
+        $this->dbConnection->update('movie', [
+            'poster_path' => $posterPath
+        ], ['id' => $id]);
+    }    
+
+    public function updateTmdbPosterPath(int $id, string $posterPath) : void
+    {
+        $this->dbConnection->update('movie', [
+            'tmdb_poster_path' => $posterPath
```
leepeuker, line 1067: currently we have this workflow:
- tmdb sync checks what the current default poster is and compares it to the `tmdb_poster_path`
- if the paths are different, the `tmdb_poster_path` will be updated with the new data
- this will trigger an image caching job, which will update the value at `poster_path` again

That means any changes you save here will probably be overwritten later again :thinking: 

I think we could add a new column `lock_image` and set set it to true if a custom images was chosen. If a movie has a custom image, than we do not run cache updates for it -> no overwrites

templates/component/modal-change-movie-poster.html.twig
```diff
@@ -0,0 +1,35 @@
+
+<div class="modal fade" id="changePosterModal" tabindex="-1" aria-labelledby="changePosterModal" aria-hidden="true" data-bs-keyboard="false">
+    <div class="modal-dialog modal-dialog-centered modal-dialog-scrollable">
+        <div class="modal-content">
+            <div class="modal-header">
+                <h1 class="modal-title fs-5" id="changePosterModalHeader"></h1>
+                <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
+            </div>
+            <div class="modal-body">
```
leepeuker, line 9: the ui looks a bit off to me. I think "Change Poster" should be enough as the the modal title and the button layout needs a bit of improvement

image

leepeuker: oh and the update poster button should be disabled I think, as long as no option was selected

src/Domain/Movie/MovieRepository.php
leepeuker: hm, later we want to probably add the option to edit all attributes of a movie, not sure of it would already make sense to think about what the best locking mechanism would be on a bigger scale than only the image. having lock columns for every attribute seems not very optimal to me

src/Api/Tmdb/TmdbApi.php
JVT038: Yes, it is possible with the `language` query parameter, but how would we determine which language should be filtered for the user? Using the country setting or..?

src/Domain/Movie/MovieRepository.php
JVT038: So I have tested it and in my environment the custom selected poster is NOT overwritten by the `tmdb:imageCache:refresh` command, so I'm not sure if we need that extra column.

Additionally, I suggest a column that's something like `status` and it has the value of either `unsynced`, `synced` or `customized`. If the item is either synced or customized, the tmdb sync check will skip this item.

public/js/movie.js
JVT038: As far as I know, TMDB images don't really have rate limiting, so I'm not sure if this is really necessary tbh.

src/Api/Tmdb/TmdbApi.php
leepeuker: Hm, good question. Maybe we could add a drop-down to select the country to the image select modal and set it default to the users country setting or eng if not set? 🤔

public/js/movie.js
leepeuker: You are right, here more about the rate limits: https://www.themoviedb.org/talk/6422345a6a34480112ba52fd

I still think that it is not a good practice to load so many images at once. If your internet connection is slow it could take ages to load and the images at the bottom can finish loading before the one at the top, which is not a good user experience.

leepeuker: You do not have to implement this, I can do this later too

JVT038: Alright, then I'll 'resolve' this conversation for now if you can do it later.

JVT038:
Can we merge this, or would you like to change some stuff? @leepeuker

leepeuker:
I try to find time this week to review and test this PR again so that we can merge it @JVT038

PR #542, opened 2023-11-05
Rework docker image

leepeuker:


docker-compose.yml
```diff
@@ -4,9 +4,8 @@ services:
   app:
     build:
       context: ./
-      dockerfile: ./build/php/Dockerfile
+      dockerfile: ./build/Dockerfile
       target: development
-    user: "${USER_ID}:${USER_ID}"
```
leepeuker, line 9: when `USER_ID` is changed the container does not work anymore because e.g. supervisor cannot be started.

PR #553, opened 2023-12-07
Remove invalid dashboard links for most watched genres and languages

leepeuker:


sahinakkaya:
Why is this change required? I remember I could click on a production company name and I could see all the movies from that production company. It was working for me. 🤔

PR #575, opened 2024-01-26
Add API endpoint for retreiving an auth token

JVT038:
This PR adds ``/authentication/request-token` where an HTTP POST request should be sent.

It also adds a new header, `X-Movary-Client`, which allows the backend to distinguish the default Movary frontend from any 3rd-party apps that may want to use Movary's authentication.

This is important, because if it's just a 3rd-party, Movary will only return the token in JSON format while if it's the Movary frontend, the backend will return a HTTP 303 response. I stole this idea from the Plex API, which does the same, only with `X-Plex-Product` instead. 

Additionally, I have added two columns to the `user_auth_tokens` table containing the Device Name (aka the value from `X-Movary-Client`) and the user agent string from the HTTP request. This allows us to add a new feature in the future, where the users can see all the tokens /  devices and invalidate any token / authorized sessions. The user can see which token belongs to which app and which user agent was used to request the token. Maybe we should store the IP as well? And maybe another column containing the date that the token was last used.

Anyway, the default client name for the Movary frontend is 'Movary Web'.

The login process for the current Twig frontend has already been migrated from the old web endpoint to the API. It no longer works with the POST form, but instead JS is used to send the POST request; this is to allow a smooth transition to a TOTP form if it's necessary.

Also, the verification of the TOTP code has been moved from the `TwoFactorAuthenticationController` to the new API `AuthenticationController`. I did this because now the user can log in with the TOTP code with one HTTP request, instead of two different requests. The 2FA controller is still used for enabling / disabling the 2FA though.

### DIFF
--- a/src/HttpController/Api/AuthenticationController.php
+++ b/src/HttpController/Api/AuthenticationController.php
@@ -1,0 +1,66 @@
+<?php declare(strict_types=1);
+
+namespace Movary\HttpController\Api;
+
+use Movary\Domain\User\Exception\InvalidCredentials;
+use Movary\Domain\User\Exception\InvalidTotpCode;
+use Movary\Domain\User\Exception\NoVerificationCode;
+use Movary\Domain\User\Service\Authentication;
+use Movary\Domain\User\Service\TwoFactorAuthenticationApi;
+use Movary\Util\Json;
+use Movary\Util\SessionWrapper;
+use Movary\ValueObject\Http\Request;
+use Movary\ValueObject\Http\Response;
+
+class AuthenticationController
+{
+    public function __construct(
+        private readonly Authentication $authenticationService,
+        private readonly TwoFactorAuthenticationApi $twoFactorAuthenticationApi,
+        private readonly SessionWrapper $sessionWrapper,
+    ) {
+    }
+
+    public function requestToken(Request $request) : Response
+    {
+        $postParameters = Json::decode($request->getBody());
+        $headers = $request->getHeaders();
+        if($postParameters['email'] === null || $postParameters['password'] === null) {
+            return Response::createBadRequest('Username or password has not been provided');
+        }
+        $totpCode = $postParameters['totpCode'] ?? 0;
+        
+        if(isset($headers['X-Movary-Client']) === false) {
+            return Response::createBadRequest();
+        }
+        
+        $client = $headers['X-Movary-Client'];
+
+        try {
+            $this->authenticationService->login($postParameters['email'], $postParameters['password'], false, (int)$totpCode);
+
+            if($client === 'Movary Web') {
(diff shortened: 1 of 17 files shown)


JVT038 (commented):
Also a weird thing I noticed is that swagger doesn't work on my local PC.

In the Network tab, I see that Swagger tries to connect with `http://movary.local/docs/movary.local/api/openapi` while it should be `http://movary.local/api/openapi`. The variable `openApiJsonUrl` is set to `"movary.local/api/openapi"` in the `OpenApiController`, but Swagger apparently interprets it wrong or something and tries to append that to the current host.

If I change the URL simply to `/api/openapi`, then it does work.

db/migrations/sqlite/20240126110433_AddDeviceNameAndUserAgentToAuthTokenTable.php
```diff
@@ -41,7 +41,6 @@ public function up() : void
             )
             SQL,
         );
-        $this->execute('INSERT INTO `user_auth_token_tmp` (`id`, `user_id`, `token`, `expiration_date`, `created_at`) SELECT * FROM `user_auth_token`');
```
leepeuker: If there are existing auth tokens the migration will fail, because the new columns device_name and user_agent are required. I think it is best to purge all existing auth tokens during the migration to prevent issues or having to use empty strings as fallback values.

src/HttpController/Api/AuthenticationController.php
```diff
@@ -0,0 +1,80 @@
...
+                $tokenRequestBody['email'],
+                $tokenRequestBody['password'],
+                (bool)$rememberMe,
+                $requestClient,
+                $request->getUserAgent(),
+                (int)$totpCode,
+            );
+        } catch (InvalidCredentials $e) {
+            return Response::createBadRequest(Json::encode([
+                'error' => basename(str_replace('\\', '/', get_class($e))),
```
leepeuker, line 50: This not really a good idea, because we tell the user exactly what went wrong which we do not want in some cases, e.g. we do not want to tell a bad actor that an email exists but the password is wrong. 
I will adjust this to only return a generic `Invalid credentials` again and only be a concrete error for ttop issues.

src/HttpController/Api/AuthenticationController.php
```diff
@@ -0,0 +1,92 @@
...
+            return Response::createUnauthorized(
+                Json::encode([
+                    'error' => 'InvalidCredentials',
+                    'message' => 'Invalid credentials'
+                ]),
+                [Header::createContentTypeJson()],
+            );
+        }
+
+        return Response::createJson(
```
leepeuker, line 85: **info**: the authenication endpoint always returns the same response on success regardless of the client

leepeuker (approved):

PR #581, opened 2024-02-17
Add support for preflight requests with CORS

JVT038:
While working on the new frontend, I encountered a new problem.

The Movary backend runs on a different address than the frontend. The frontend sends HTTP requests to a different address and due this, we need a proper Cross-Origin-Resource-Sharing (CORS) policy. More info [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS).

To 'discover' the CORS policy, the browser sends a [preflight](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS) request. This request is basically asking what methods are allowed for the API endpoint and this PR adds a proper response to the request.

### DIFF
--- a/public/index.php
+++ b/public/index.php
@@ -5,6 +5,7 @@ session_start();
 /** @var DI\Container $container */
 
 use Movary\HttpController\Web\ErrorController;
+use Movary\HttpController\Api\PreflightRequestController;
 use Movary\ValueObject\Http\Request;
 use Movary\ValueObject\Http\Response;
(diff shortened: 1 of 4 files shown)


JVT038:
I'm closing this, because I think it's better to let Vite act as a [proxy](https://vitejs.dev/config/server-options.html#server-proxy).

All the requests to `http://frontend/api` will be redirected to `http://backend/api`.

If the proxy doesn't turn out to be a good solution after all, we can reopen the PR.

PR #582, opened 2024-02-19
Add authentication check endpoint

JVT038:
This PR adds an API endpoint to check if the user is authenticated or not.
The path is `/authentication/authenticated` and it returns a JSON object with `authenticated = false || true`.

If the user is authenticated, the object will also include a `username` and `userId` property, with the values of the current username and ID. 

This PR is based on #575 , so it has to be merged after that one.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -1066,13 +1066,13 @@
         }
       }
     },
-    "/authentication/authenticated": {
+    "/authentication/token": {
       "get": {
         "tags": [
           "Authentication"
         ],
-        "summary": "Checks the authentication status of the user.",
-        "description": "Checks if the user is authenticated based on cookie.. The response is a JSON object with the authentication status and an optional username + userId of the authenticated user",
+        "summary": "Retrieves the authhentication data",
+        "description": "Gets the user ID, the username and whether the user is admin or not, based on a token in either a header or a cookie.",
         "responses": {
           "200": {
             "description": "OK",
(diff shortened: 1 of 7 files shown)


JVT038:
BTW, the reason that I think this endpoint is necesssary, is because there's currently no way for the new Vue frontend to figure out whether the user is authenticated or not. Additionally, if the user was authenticated, the frontend still can't figure out which user is authenticated.

docs/openapi.json
```diff
@@ -1066,6 +1066,45 @@
         }
       }
     },
+    "/authentication/authenticated": {
```
leepeuker, line 1069: **todo**: needs to be adusted, as well as the create token response

settings/routes.php
```diff
@@ -202,6 +202,7 @@ function addApiRoutes(RouterService $routerService, FastRoute\RouteCollector $ro
     $routes->add('GET', '/openapi', [Api\OpenApiController::class, 'getSchema']);
     $routes->add('POST', '/authentication/token', [Api\AuthenticationController::class, 'createToken']);
     $routes->add('DELETE', '/authentication/token', [Api\AuthenticationController::class, 'destroyToken']);
+    $routes->add('GET', '/authentication/token', [Api\AuthenticationController::class, 'getTokenData']);
```
leepeuker, line 205: **info**:  my suggestion for the endpoint to stay consistent with the other endpoints

src/Domain/User/Service/Authentication.php
```diff
@@ -98,14 +98,19 @@ public function getCurrentUserId() : int
         return $userId;
     }
 
-    public function getToken() : ?string
+    public function getToken(Request $request) : ?string
```
leepeuker, line 101: **info**: we should have only one place where we extract the auth token from the request

src/HttpController/Api/AuthenticationController.php
```diff
@@ -85,8 +86,12 @@ public function createToken(Request $request) : Response
 
         return Response::createJson(
             Json::encode([
-                'userId' => $userAndAuthToken['user']->getId(),
-                'authToken' => $userAndAuthToken['token']
+                'authToken' => $userAndAuthToken['token'],
+                'user' => [
```
leepeuker, line 90: **info**: I though it would make sense for the create token and get token endpoints to return the same user data

settings/phpcs.xml
```diff
@@ -18,7 +18,11 @@
     <rule ref="Generic.Commenting.Fixme"/>
     <rule ref="Generic.ControlStructures.InlineControlStructure"/>
     <rule ref="Generic.Files.ByteOrderMark"/>
-    <rule ref="Generic.Files.LineEndings"/>
+    <rule ref="Generic.Files.LineEndings">
+        <rule ref="Generic.Files.LineEndings">
+            <exclude name="Generic.Files.LineEndings.InvalidEOLChar"/>
+        </rule>
+    </rule>
```
JVT038, line 25: I have changed this, because code sniffer leaves an invalid line ending error due to my Windows installation. See [here](https://github.com/squizlabs/PHP_CodeSniffer/issues/2554)

src/HttpController/Api/AuthenticationController.php
```diff
@@ -114,4 +119,37 @@ public function destroyToken(Request $request) : Response
...
+                [Header::createContentTypeJson()],
+            );
+        }
+
+        $user = $this->userApi->findByToken($token);
+        if ($user === null) {
+            return Response::createUnauthorized();
+        }
+
+        if($this->authenticationService->isUserAuthenticatedWithCookie() && $this->authenticationService->isValidAuthToken($token) === false) {
```
leepeuker, line 141: what if a user is using the auth token, we do not test in this case if the token is expired?
let us fix this in the followup PR

leepeuker (approved):

PR #584, opened 2024-02-22
Add API endpoint for destroying tokens

JVT038:
This PR adds an API endpoint that can destroy API tokens (manually generated by the user in the settings) and authentication tokens (generated upon login of the user).

Also important to mention that this PR is based on #575 , so that PR has to be merged before this one.

### DIFF
--- a/public/js/app.js
+++ b/public/js/app.js
@@ -498,3 +498,11 @@ function addAlert(parentDivId, message, color, withCloseButton = true, marginBot
 function removeAlert(parentDivId) {
     document.getElementById(parentDivId).innerHTML = ''
 }
+
+async function logout() {
+    await fetch('/api/authentication/token', {
+        method: 'DELETE',
+    });
+
+    window.location.href = '/'
+}
--- a/public/js/login.js
+++ b/public/js/login.js
@@ -11,6 +11,11 @@ async function submitCredentials() {
         return;
     }
 
+    const forbiddenPageAlert = document.getElementById('forbiddenPageAlert');
+    if (forbiddenPageAlert) {
+        forbiddenPageAlert.classList.add('d-none');
+    }
+
     if (response.status === 400) {
         const error = await response.json();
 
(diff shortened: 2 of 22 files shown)


leepeuker:
I have removed the `IsAuthenticated` middleware check from the new endpoint, I think we should not tell a user if a auth token really existed, only if the deletion process run without an error. Additionally, the `IsAuthenticated` middleware should in theory fail if the token is expired, which means you could not delete an expired token which seems not really practical.
Instead we will no return a 400 error if the Auth token header is missing.

public/js/login.js
```diff
@@ -11,6 +11,11 @@ async function submitCredentials() {
         return;
     }
 
+    const forbiddenPageAlert = document.getElementById('forbiddenPageAlert');
```
leepeuker, line 14: **info**: while testing the logout I noticed that the `Sign in to access page` did not vanish when other error messages are displayed (e.g. `Invalid credentials`). Displaying two alerts seems weird, so I added this

public/js/app.js
```diff
@@ -498,3 +498,11 @@ function addAlert(parentDivId, message, color, withCloseButton = true, marginBot
 function removeAlert(parentDivId) {
     document.getElementById(parentDivId).innerHTML = ''
 }
+
+async function logout() {
+    await fetch('/api/authentication/token', {
+        method: 'DELETE',
+    });
+
+    window.location.href = '/'
```
leepeuker, line 507: **info**: I only wanted to reload the current page initially on logout, so that the user is only redirect to the login page if he currently was on a protected page, but that this would display the `Sign in to access page` alert, which seems weird if the user purposefully logged out, so I decided for this way

tests/rest/api/authentication.assert.http
```diff
@@ -0,0 +1,97 @@
+POST http://127.0.0.1/api/authentication/token
```
leepeuker, line 1: This is interesting feature in phpstorm, you can run assertions against the http requests.
To execute every request in error press the green "Run all requests" button:
image

You can see the result in the `Run` window and which (if any) assertion failed, example with a failed assertion:
image

This feature is a bit bulky to use, but at least we both can run the same tests in an automatic way locally.

leepeuker (dismissed):

leepeuker (approved):

PR #589, opened 2024-02-25
Add API endpoint for images

JVT038:
I have added two API endpoints:

- `/api/images/movies/{id}`
- `/api/images/person/{id}`

They return a 404 if the ID doesn't exist or if the item doesn't have an image. If the image does exist, then it returns 200 with the raw image data attached in the body.


JVT038:
@leepeuker What do you think about adding something like imagejpeg or ImageMagick to compress the returned images?

JVT038:
So I merged main into this branch, and this got me thinking: Do we even need this?

Can't we just load the images from TMDB themselves instead of doing it like this?

Or maybe we can use this API endpoint to act as a 'proxy'. Then the client will send an HTTP GET request for an image and this endpoint will either return the image from its own cache or proxy the request to the TMDB servers if the image isn't cached. Although for this functionality, the PR will have to be overhauled.

PR #590, opened 2024-02-26
Add API endpoint to create new user

JVT038:
Besides creating a new API endpoint in `/api/create-user`, I have also refactored some stuff related to the OpenAPI specs and the middleware of creating a new user.


JVT038:
Idk why, but this PR is apparently partially based on #583 and #585, so check those first before doing this one.

PR #593, opened 2024-02-27
Change token header name

JVT038:
It has been changed in all occurrences to `X-Movary-Token` (with the CTRL + SHIFT + H shortcut) and while I haven't tested it yet, it should be good in theory.

Besides that, I have decided to move all tests in `authentication.assert.http` to `authentication.http`, because the tests in `authentication.http` were basically the same, but without the actual testing / assertion statements.


JVT038:
If we update PHP Code Sniffer, we can use readonly classes, because they [fixed it](https://github.com/squizlabs/PHP_CodeSniffer/issues/3727) in the newer versions.

leepeuker (approved):

PR #594, opened 2024-02-27
Update dependencies

JVT038:
All the tests, unit tests and HTTP tests in the `main` branch still work properly.


JVT038:
If we update PHP Code Sniffer, we can use readonly classes, because they https://github.com/squizlabs/PHP_CodeSniffer/issues/3727 in the newer versions.

leepeuker (approved):

leepeuker:
> If we update PHP Code Sniffer, we can use readonly classes, because they [fixed it](https://github.com/squizlabs/PHP_CodeSniffer/issues/3727) in the newer versions.

The code sniffer project we use is abandoned by the maintainer, we should probably switch to https://github.com/PHPCSStandards/PHP_CodeSniffer/

Issue #572, opened 2024-01-23
To-do list for the new frontend


JVT038:
## Frontend (the new repo)

URL to the new repo: https://github.com/leepeuker/movary-frontend

### Movie stuff 
- [ ] Add the dashboard page
- [ ] Add the all movies page
- [ ] Add the movie details page
- [ ] Add the person details page
- [ ] Add the watchlist page
- [ ] Add the 'Add movie' modal
- [ ] Add the top watched directors page
- [ ] Add the top watched actors page

### Settings stuff
- [ ] Add the account settings -> general page
- [ ] Add the account settings -> security page
- [ ] Add the account settings -> dashboard page
- [ ] Add the account settings -> personal data page
- [ ] Add the integrations settings -> Plex page
- [ ] Add the integrations settings -> Jellyfin page
- [ ] Add the integrations settings -> Emby page
- [ ] Add the integrations settings -> Radarr page
- [ ] Add the integrations settings -> Trakt.Tv page
- [ ] Add the integrations settings -> Letterboxd page
- [ ] Add the integrations settings -> Netflix page
- [ ] Add the server settings -> general page
- [ ] Add the server settings -> users page
- [ ] Add the server settings -> Email page
- [ ] Add the server settings -> Jobs page
- [ ] Add the settings -> app page

### New features
Why not improve the frontend when we're already overhauling it?

- [ ] Add a new feature in the settings -> job page where users can manually add  / trigger a new job (such as syncing the TMDB data)
- [ ] Add a new feature where users can manually edit the cover of the movie (this would override #540 and close that PR in favour of the new frontend. Also, it would close #400)
- [ ] Add a new feature where users can manually edit the image of a person
- [ ] Add a backdrop / background to the movie details page (closes #248)
- [ ] Add an annual recap that triggers in December (closes #466)
- [ ] Add a feature to save the sorting options (closes #441)
- [ ] Improve the dashboard rows performance by lazy-loading them (closes #394)
- [ ] Make the person pages more detailed (closes #383)
- [ ] Add support for translations of the web UI (probably with something like [I18Next](https://www.i18next.com); partially closes #577)
- [ ] Add shortcut to open 'log movie' modal and immediately focus on the input (closes #571)

### Fixing bugs
- [ ] Fix dashboard item list bottom corners (closes #330)

### Other stuff
- [ ] Add dark mode
- [ ] Add the Swagger API UI

## Backend (this repo)
[...]

PR #595, opened 2024-02-28
Added endpoint to retrieve movie data

JVT038:
I have also moved the `Api\MovieSearchController` to `Api\MovieController`, because I thought that made more sense for the movie stuff to be in the same controller.

I have added HTTP tests for the new endpoints and I have added the new endpoint in the openapi specs.

I decided to just copy and paste all the data that's shown in the movie page into this endpoint response, because this endpoint will be used by the new frontend to render the movie page.

Part of #572


JVT038:
I merged #598 into this one, because otherwise API authentication doesn't work. So first merge/review #598 before moving onto this one.

PR #614, opened 2024-09-01
Add option to dis/enable the locations feature

leepeuker:


leepeuker:
Todo:
- integrate endpoints with the frontend
- hide locations input if feature is disabled

PR #623, opened 2024-12-18
Automatically run migrations on start

### LINKED ISSUE #342: Run migrations automatically when a Docker container starts
Just add some commands to the Docker script to check for new migrations and if there are any, execute them.
This will prevent issues where users forget to run migrations on new versions.

TODO:
- [ ] Add option to run migrations automatically on container start (set via env variable?)
- [ ] Update docs and explain how this works at https://docs.movary.org/install/docker/

JVT038:
Added supervisord sections for migrations, set to run once.

I have not tested this yet, although it should work in theory. 

Fixes #342


settings/supervisor/database-migrate.conf
```diff
@@ -0,0 +1,14 @@
+[program:movay-database-migrate]
+command=bash -c 'if [ "$DISABLE_DATABASE_AUTO_MIGRATION" != "true" ]; then /usr/local/bin/php /app/bin/console.php database:migration:migrate; else echo "Database migration is disabled"; fi'
```
leepeuker, line 2:
```suggestion
command=bash -c 'if [ "$DATABASE_DISABLE_AUTO_MIGRATION" != "true" ]; then /usr/local/bin/php /app/bin/console.php database:migration:migrate; else echo "Database migration is disabled"; fi'
```

leepeuker (approved):

PR #624, opened 2024-12-19
Refactor supervisor config

leepeuker:


docker-compose.yml
```diff
@@ -9,6 +9,7 @@ services:
       - "${HTTP_PORT}:80"
     volumes:
       - ./:/app
+      - ./settings/supervisor:/opt/docker/etc/supervisor.d
```
leepeuker, line 12: run supervisor in local development environment

settings/supervisor/job-processing.conf
```diff
@@ -1,4 +1,4 @@
-[program:movary]
+[program:movay-job-processing]
```
leepeuker, line 1: let us split the supervisor processes into separate files, makes it easier to manage imo

settings/supervisor/database-migrate.conf
```diff
@@ -0,0 +1,13 @@
+[program:movay-database-migrate]
+command=/usr/local/bin/php /app/bin/console.php database:migration:migrate
+numprocs=1
+user=application
+autostart=true
+autorestart=unexpected
+exitcodes=0
+startsecs=0
```
leepeuker, line 8: we must not require a min runtime, my last migration took `0.0140` seconds

settings/supervisor/database-migrate.conf
```diff
@@ -0,0 +1,13 @@
+[program:movay-database-migrate]
+command=/usr/local/bin/php /app/bin/console.php database:migration:migrate
+numprocs=1
+user=application
+autostart=true
+autorestart=unexpected
+exitcodes=0
```
leepeuker, line 7:
```suggestion
autorestart=unexpected
exitcodes=0
```

retry if the expected status code was not returned (maybe same initial connection issues?)

docker-compose.yml
```diff
@@ -9,6 +9,7 @@ services:
       - "${HTTP_PORT}:80"
     volumes:
       - ./:/app
+      # - ./settings/supervisor:/opt/docker/etc/supervisor.d
```
leepeuker, line 12: example how to run supervisor in local development environment

leepeuker:
@JVT038 I made a few adjustments extending your proposal. what do you think?

docs/configuration.md
```diff
@@ -32,16 +32,17 @@ The `Web UI` column is set to yes if an environment variable can alternatively b
...
+| `DATABASE_MODE`                   |            -            | **Required** `sqlite` or `mysql`                       |
+| `DATABASE_SQLITE`                 | `storage/movary.sqlite` |                                                        |
+| `DATABASE_MYSQL_HOST`             |            -            | Required when mode is `mysql`                          |
+| `DATABASE_MYSQL_PORT`             |         `3306`          |                                                        |
+| `DATABASE_MYSQL_NAME`             |            -            | Required when mode is `mysql`                          |
+| `DATABASE_MYSQL_USER`             |            -            | Required when mode is `mysql`                          |
+| `DATABASE_MYSQL_PASSWORD`         |            -            | Required when mode is `mysql`                          |
+| `DATABASE_MYSQL_CHARSET`          |        `utf8mb4`        |                                                        |
+| `DATABASE_DISABLE_AUTO_MIGRATION` |           `0`           | On default docker runs migrations on container startup |
```
leepeuker, line 45: this is the new addition

leepeuker:
> I looked at your changes and they seem fine to me.
> 
> The only thing I'm concerned about is the order of the supervisor processes; If there's a job schedules which depends on a migration, then we have to make sure that the migration is ran before the job processor starts running.
> 
> I just pushed a commit which sets the priority and ensures that the migrations are executed before the jobs.

hm, the priority only decides the order of starting the processes, so this does not really solve the issue in general, especially if the migrations takes a few seconds or longer. but I do not think we can solve this issue really via supervisor, this is not what supervisor was meant to do.

JVT038:
You might be right, but I'm pretty sure that if a job fails, the job will just be restarted 15 seconds later right? So we can also just keep it the way it is, and hopefully no issues regarding the process order will occur lol.

JVT038:
I think we should just YOLO merge this and we'll eventually find out if something is wrong lol

PR #629, opened 2025-02-01
Refactor docker (compose) setup

leepeuker:


docs/development/setup.md
```diff
@@ -1,27 +1,30 @@
...
-- run `make reup` to stop and restart all containers
-- run `make app_database_migrate` execute the database migrations
-- run `make app_jobs_process` process the next job from the queue (see database table `job_queue`)
+- `make up` to start the application in production stage using SQLite (using docker volumes)
+- `make up_mysql` to start the application in production stage using MySQL (using docker volumes)
+- `make up_development` to start the application in development stage using SQLite (mounting local files)
+- `make up_development_myqsl` to start the application in development stage using MySQL (mounting local files)
+- `make down` to stop all containers
+- `make app_database_migrate` execute the database migrations
+- `make app_jobs_process` process the next job from the queue (see database table `job_queue`)
```
JVT038, line 27: Maybe add that make only works on linux.

leepeuker: Make maybe only works on linux but the commands executed the make targets should work on all systems, so the make targets can still provide a good starting point to see how to run the application. We should definitely add this to the docs :+1:

Issue #631, opened 2025-02-13
won't start on Synology NAS


rastarr:
### Description

this is driving me crazy. Tried both mysql and now just trying SQLite

My docker compose is:
```
version: "3.2"

services:
  movary:
    image: leepeuker/movary:latest
    container_name: movary
    restart: "no"
    ports: 
      - "8712:80"
    environment:
      TMDB_API_KEY: "<redacted>"
      TMDB_ENABLE_IMAGE_CACHING: 1
      TIMEZONE: "Asia/Bangkok"
      DATABASE_MODE: "sqlite"

    volumes:
      - movary-storage:/app/storage
volumes:
  movary-storage:
```

And the logs suggest some binding issue, I think with nginx, in some way:
```
INFO: Automatic database migration is enabled
using config file settings/phinx.php
using config parser php
using migration paths 
 - /app/db/migrations/sqlite
warning no environment specified, defaulting to: dynamic
using adapter sqlite
using database storage/movary.sqlite
ordering by creation time
All Done. Took 0.0163s
SUCCESS: Automatic database migration succeeded
INFO: Generating symbolic link for storage
2025-02-13 02:54:17:318908 - Public storage symlink created.
2025-02-13 02:54:18,174 INFO supervisord started with pid 1
2025-02-13 02:54:19,177 INFO spawned: 'kill-supervisor-on-error' with pid 12
2025-02-13 02:54:19,179 INFO spawned: 'php-fpm' with pid 13
2025-02-13 02:54:19,180 INFO spawned: 'nginx' with pid 14
2025-02-13 02:54:19,183 INFO spawned: 'movary-job-processing' with pid 15
2025/02/13 02:54:19 [emerg] 14#14: bind() to [::]:80 failed (13: Permission denied)
2025-02-13 02:54:19,243 WARN exited: nginx (exit status 1; not expected)
2025-02-13 02:54:19,272 INFO gave up: nginx entered FATAL state, too many start retries too quickly
[13-Feb-2025 02:54:19] NOTICE: fpm is running, pid 13
[13-Feb-2025 02:54:19] NOTICE: ready to handle connections
2025-02-13 02:54:19:313019 - Processing job...
2025-02-13 02:54:19:314865 - Nothing to process.
2025-02-13 02:54:19:314926 - Sleeping for 14.998081 seconds to reach min runtime...
[...]

PR #633, opened 2025-02-16
Change http port from 80 to 8080

leepeuker:


JVT038:
Why?

leepeuker:
> Why?

See the related issue: https://github.com/leepeuker/movary/issues/631

PR #639, opened 2025-05-18
Build correct radarr API URL

alexcaza:
The previous implementation was using the webhook URL builder, which was building an incorrect path: `/api/webhook/api/feed/radarr`.

The path should be `/api/feed/radarr`.

Let me know if you'd prefer a different implementation. Seeing as there's only two places the URL is build (the router and then in here), duplication seems fine. Building an abstraction for this URL seems overkill.

### DIFF
--- a/src/Service/WebhookUrlBuilder.php
+++ b/src/Service/WebhookUrlBuilder.php
@@ -23,11 +23,6 @@ class WebhookUrlBuilder
         return $this->buildUrl('plex', $webhookId);
     }
 
-    public function buildRadarrFeedUrl(string $feedId) : ?string
-    {
-        return $this->buildUrl('api/feed/radarr', $feedId);
-    }
-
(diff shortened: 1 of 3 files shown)


src/Service/WebhookUrlBuilder.php
```diff
@@ -23,11 +23,6 @@ public function buildPlexWebhookUrl(string $webhookId) : ?string
         return $this->buildUrl('plex', $webhookId);
     }
 
-    public function buildRadarrFeedUrl(string $feedId) : ?string
```
leepeuker, line 26: This method is used in `SettingsController` as well (line 390), we have to adjust this place as well

alexcaza: Sorry about that! I thought I searched project wide, but I guess I forgot 🙃

PR #677, opened 2025-08-18
Add/search movie description

alifeee:
this replaces #672 (local branch PR instead of remote branch PR)

a small change, as it seemed to make search results more clear

### before

which Closure is which? who is to say

image

### after

they have descriptions now. very nice

image

### notes

a few "magic numbers" were picked (i.e., descriptions truncated to 100 characters) — feel free to change/export these

everything (styles) is done in JS as there is no CSS stylesheet that is loaded on every page (but the "search box" is available on every page) — not being sure of the layout of the project, I am not sure a better way to do this (styling). maybe with bootstrap classes.


alifeee:
> I think `...` should not be added if there is nothing cut of :thinking: 

I believe a17d0ba7aefd0da528b1393d9fcc4ba6d6c03c1a will have fixed that

leepeuker:
Okay I tested it and it looks great, thank you!

leepeuker (approved):

Issue #676, opened 2025-08-18
Add movie names (especially for movies with missing posters)


alifeee:
hi :]

the problem here is this: if a movie (or multiple) have no posters (or bad posters), it is very difficult to see what they are. For example, this page:

<img width="667" height="864" alt="Image" src="https://github.com/user-attachments/assets/f2417ee2-d09c-4fbe-8b78-ea7e76740964" />

I propose two options:

## option 1 – title all movies

see branch https://github.com/leepeuker/movary/tree/add/movie-names

this is a simple addition, but changes the "look and feel" of all the blocks of movies, so you may not like it so much. I copied the HTML from the actors/directors section and added it to the movies. This solves the problem as you can see titles for every movie, so the poster is no longer the most important piece.

<img width="667" height="864" alt="Image" src="https://github.com/user-attachments/assets/381c543c-c0f3-44a4-8555-57b0d7601127" />

## option 2 - add title to missing movie posters

see branch https://github.com/leepeuker/movary/tree/add/missing-movie-names

the branch is less complete here, I was just toying with the idea. But in effect, we mark movies without a poster with a flag, then can use this flag in `twig` to switch between showing a poster or showing a poster-sized name

<img width="667" height="864" alt="Image" src="https://github.com/user-attachments/assets/0aa47e12-ba07-4bbc-88a9-049fca8f5a55" />

## notes

option 2 is probably my favourite, as it doesn't change the "look and feel" of the movie boards so much, but makes them a lot more useful. I don't much like the font size/font choice that I have picked, but it is just as a demonstration for now and could be changed

what do you think @leepeuker (or anyone else?)

alifeee:
sure, I think option 2 makes most sense then. I have changed the styles so it looks like this:

<img width="966" height="609" alt="Image" src="https://github.com/user-attachments/assets/becf78a0-492d-4d53-a747-102bfbdb89fe" />

…which is quite similar to the "No Image" text (ish — as close as we can get with web-safe fonts)

what do you think of how the backend is done here? https://github.com/leepeuker/movary/compare/main..add/missing-movie-names (in [src/Service/ImageUrlService.php](https://github.com/leepeuker/movary/compare/main..add/missing-movie-names#diff-c0c7e89e774452bec02aec0dad888fef1103bc6c421e2c294d821a24f4f25c81) — I'd like it to be a way that you think is good before adding it to any other pages that currently show "No Image"

leepeuker:
[...]

PR #678, opened 2025-08-18
fix hyperlinking

alifeee:
a lot of the front pages use this style of "link":

```html
<div onclick="window.location='{{ applicationUrl }}/users/{{ routeUsername }}/persons/{{ mostWatchedActor.id }}'"
…
</div>
```

I propose that this is bad semantic HTML and should be replaced by

```html
<a href="{{ applicationUrl }}/users/{{ routeUsername }}/persons/{{ mostWatchedActor.id }}"
…
```

## why

the personal reasons that I think this is better include:

- you can hover over cards and see the URL (especially useful without posters – see also #676)
- you can right click/long press cards and open in new tab
- you can middle click cards and open in new tab
- (generally) there is more browser support for "links" when they are "links" and not JS hacks

for some external motivation, [this SO post](https://stackoverflow.com/questions/1667416/window-location-href-vs-clicking-on-an-anchor) suggests:

> 
> 
> Wherever possible, you should use <a href="foo.html"> over window.location.href, for a number of very good reasons.
> 
> 1. If you have javascript disabled, none of the links would work.  
> 1. Spiders, such as Google Bot, do not interpret javascript, and so they won't follow any of your links.  
> 1. IT BREAKS THE INTERNET. No, really though - the World Wide Web is built on the very basis of discoverable linkages between pages. Hiding these linkages with non-standard .. err, links, goes against that very premise.  
> 1. It makes for a bad user experience: a user expects that when they mouse over a link, they will have access to some information:
>     1. the destination displayed in the status bar (very important!)
>     1. right-click -> copy link location
>     1. middle-click -> open new tab
>     1. etc
>     1. Using window.location breaks all of these

see here that when I hover the card it shows the URL in the bottom left (this is not currently the case – it shows nothing)

image

## notes

I do not think I have hit every example here. For example the watchlist page shows a modal on click, so would need the behaviour changing to change to an anchor `<a>` tag (I think probably this could be removed and you could just leave the "remove from watchlist" on the movie page – though this is a separate concern.)

image

in that vein, this may not be a complete PR, and requires a little more editing. I would also love feedback from @leepeuker or someone else about this

### DIFF
--- a/templates/page/movie.html.twig
+++ b/templates/page/movie.html.twig
@@ -115,17 +115,6 @@
         {# --- Buttons --- #}
 
         <div class="container text-center" id="toggleWatchDatesButtonDiv">
-            {% if isOnWatchlist %}
-            <div class="row mb-2">
-                <a
-                    href="{{ applicationUrl }}/users/{{ routeUsername ?: currentUserName }}/watchlist"
-                    class="col-lg-5 col-xl-4 col-xxl-3 btn bg-primary text-light"
-                    id="onWatchListNote"
-                >
-                    On Watchlist
-                </a>
-            </div>
-            {% endif %}
             <div class="row mb-2">
                 <button type="button" class="col-lg-5 col-xl-4 col-xxl-3 btn {{ theme == 'dark' ? 'btn-light' : 'btn-dark' }}"
                         onclick="toggleWatchDates()"


alifeee:
not quite sure how to change the code to satisfy the type checker…

it complains about

```text
 ------ -------------------------------------------------------- 
  Line   src/Factory.php                                         
 ------ -------------------------------------------------------- 
  337    Parameter #3 $subject of function preg_replace expects  
         array<float|int|string>|string, string|null given.      
         🪪 argument.type                                        
 ------ --------------------------------------------------------
```

for the code

https://github.com/leepeuker/movary/blob/3a8fd9d93fde755b0c108a4e70cb874cacdc1f73/src/Factory.php#L334-L339

changing it to this doesn't fix anything

```diff
diff --git a/src/Factory.php b/src/Factory.php
index 87eec121..fcf31259 100644
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -331,10 +331,10 @@ class Factory
 
         // slugify filter for "nice looking" URLs
         //  turns names/movie titles into slugs for use in, e.g., "/…/14-freakier-friday/"
-        $twig->addFilter(new \Twig\TwigFilter('slugify', function ($string) {
+        $twig->addFilter(new \Twig\TwigFilter('slugify', function (string $string) {
-            $string = strtolower($string);
+            $string = strtolower($string) ?? "";
-            $string = preg_replace('/[^a-z-]/', "-", $string);
+            $string = preg_replace('/[^a-z-]/', "-", (string) $string) ?? "";
-            $string = preg_replace('/-+/', '-', $string);
+            $string = preg_replace('/-+/', '-', (string) $string) ?? "";
             return $string ? $string : "";
         }));
```

alifeee:
> I have fixed this by refactoring the slugify function.

thanks :] and you also moved it into somewhere in the file system I only hope to ever find out where (I am endlessly confused by Factories and Classes)

I've made a fix and a couple of changes, namely:

The default canonical URL *was* `/users/<user>/movies/9`. I made it `/users/<user>/movies/9-<movie-name-slug>`, which I was able to do thanks to that the slugify function was somewhere importable ;] (not sure I've imported it in a way you would expect, though…)

Also, made it so this entire card is clickable:

image

…which might break the bootstrap "hover" on "character name" on mobile… not sure how to test on mobile, though…

<details><summary>attempted to fix this with this change, but CSS is tricky and I have to go now</summary>

```diff
diff --git a/templates/page/movie.html.twig b/templates/page/movie.html.twig
index f82e82e9..e3f30ef4 100644
--- a/templates/page/movie.html.twig
+++ b/templates/page/movie.html.twig
@@ -208,6 +208,10 @@
         <div class="carousel">
             {% if castMembers|length == 0 %} - {% endif %}
             {% for castMember in castMembers|slice(0, 40) %}
+                <div
+                    class="card"
+                    style="width: 1rem;min-width: 110px;margin-right: 0.5rem;"
+                >
                 <a
                     class="card"
                     style="width: 1rem;min-width: 110px;margin-right: 0.5rem;"
@@ -226,6 +230,7 @@
                     <img src="{{ castMember.posterPath }}"
                          alt="{{ castMember.name }} Poster"
                          class="{% if displayCharacterNames == false %}card-img-bottom{% endif %}">
+                </a>
                     {% if displayCharacterNames == true %}
                         <span class="card-footer text-truncate"
                              id="castCharacterName"
@@ -235,7 +240,7 @@
                             <small class="text-muted" style="font-size: 0.9rem;">{{ castMember.characterName }}</small>
                         </span>
                     {% endif %}
-                </a>
+                </div>
             {% endfor %}
         </div>
```

</details>

…will look back at it soon

alifeee:
as a "simple" fix, the blue sections are `<a>` links, and the orange section is just a `<span>`.

so… the bottom should be pressable on mobile without changing the page

image

alifeee:
I think the changes here are ready. The summary is:

- allow movies and persons to have an optional "name slug" (`/12-thor-ragnarok` instead of just `/12`)
- change movies to hyperlinks `<a>` instead of using JavaScript for navigation (see above)
- add href canonical tag to pages to compensate for multiple valid URLs
- remove modal on watchlist
- rearrange buttons on `movie` page a little to compensate for the watchlist changes

I am in no rush. Do let me know what you think @leepeuker in your own time :]

leepeuker:
Pro:
- I have tested around a lot and did not notice any errors
- Good idea to move the "Remove from Watchlist" button up out of the modal
- Great to get rid of the watchlist modal :D

Con:
- I have issues with the "On Watchlist button" on the movie detail page
    - All buttons are actions directly related to the movie except the "On Watchlist" button
    - It is probably a good idea to link to the watchlist in general, but maybe not as a button like this
    - The button color looks terrible in dark mode :D
- I propose that we make only the actor posters clickable
    - You already excluded the role name, I suggest to exclude the actors name as well => this makes selecting and copy names much easier

Summary:
I think we can merge this if the exclude the new "On Watchlist button" and adjust the actors name linking.
What do you think @alifeee ?

alifeee:
small note before a proper reply: 30e2bcb59e2a25dcebda5f219b5871751253a190 seems to be all formatting changes — I have an auto-formatter, but have it disabled for Movary as it would introduce a lot of changes to git.

Do you use an auto-formatter? If so, which is it ? – so I can install it and not worry about causing formatting issues like 30e2bcb59e2a25dcebda5f219b5871751253a190

alifeee:
>     * I have issues with the "On Watchlist button" on the movie detail page
>       
>       * All buttons are actions directly related to the movie except the "On Watchlist" button
>       * It is probably a good idea to link to the watchlist in general, but maybe not as a button like this
>       * The button color looks terrible in dark mode :D

I would agree. I was struggling to find a place to put something that confirmed if a movie was on your watchlist or not. I do not think that the presence of a "remove from–" or "add to–" watchlist button is a good enough indication

Perhaps instead of the button, somewhere like here?

image

Can you think of a better place?

>     * I propose that we make only the actor posters clickable
>       * You already excluded the role name, I suggest to exclude the actors name as well => this makes selecting and copy names much easier and the UX more streamlined (two same looking fields have the same behavior)
> Edit: Same for directors!

This sounds good to me ;]

> Summary: I think we can merge this if the exclude the new "On Watchlist button" and adjust the actors name linking. What do you think @alifeee ?

I agree – I will make the necessary updates and break out the "on watchlist" notifier to a different PR… soon ;]

alifeee:
> I am using the phpstorm native auto-formatter, you can find the config here: https://github.com/leepeuker/movary/blob/main/settings/phpstorm.xml
> Sadly this is not really portable to vscode as far as I know. I am using this because it is a similiar setup to my workplace, where everyone is using phpstorm, so interoperability was not really an issue there.
> I am open to discuss switching to a IDE agnostic tooling (e.g. PHP-CS-Fixer) as this makes more sense for a public open source project, but I do not think I would find the time personally to implement this.

hm, yes. to me, it would seem the easiest option would be to change your PHPStorm to a standard PHP format. Running `phpcbf -i` on my CP results in:

```text
The installed coding standards are PSR12, MySource, PSR2, PEAR, PSR1, Squiz and Zend
```

So, if any of those are available in PHPStorm, or are already the active preset for Movary – do let me know. And if they are but Movary uses a different one, let me know if you'd be happy to switch. [PSR-12](https://www.php-fig.org/psr/psr-12/) seems to be the most popular one.

Of course, you might have created opinionated rules for yourself, which would make it harder. I do not know. ;] For now, I turn off my auto-formatter and you may have to make formatting changes to suit your current preferences. I do not have PHPStorm.

PR #684, opened 2025-08-31
add TMDB link on movie search

alifeee:
on the movie search, it would be nice to see "more info" about a movie without adding it to your watchlist or recording a play.

image

It seems fruitless to import each movie into Movary like this, so I suggest adding a TMDB link. I've copied the one from the movie page (this one)

image

## before

image

## after

image

I believe I've got the JavaScript hiding and showing done correctly (e.g., TMDB logo link has the `href` updated on new movie selected, and is hidden if the modal is opened from an individual movie page – as that page already has a TMDB link)

p.s., I am assuming from the history of the issue that #572 is semi-abandoned, so I am happy to continue making front-end changes (I think it works just fine with PHP)


leepeuker:
Good idea, worked as expected for me while testing

leepeuker (approved):

PR #685, opened 2025-09-03
Add redirect on add to watchlist

alifeee:
(note: this PR builds on #678)

the "add to watchlist" button here

image

The JavaScript code for these buttons is like this:

https://github.com/leepeuker/movary/blob/0eb6791e1c6d86f8dd01ecedbdb890d889127730/public/js/app.js#L337-L342

…i.e., it reloads the page

instead, it could redirect to the newly added movie

so instead of reloading the page, it would redirect to e.g., "/users/<user>/movies/32-anatomy-of-a-fall"

## implementation

I have added a working implementation of this in this PR, for the watchlist only.

It could:

- be on "add watch" too
- just redirect to "/movies/<id>" and have the "/movies/12" page redirect to e.g., "/movies/12-the-brasher-doubloon" — this is how TMDB does it, try loading <https://www.themoviedb.org/movie/6615> and see that you are redirected to <https://www.themoviedb.org/movie/6615-lars-and-the-real-girl> — not sure how this would fit into the PHP router, but it seems doable however it is done

as I said, this is still just in the ideas stage, but feedback is welcome before I continue to implement it :]


alifeee:
> > instead, it could redirect to the newly added movie
> 
> Hm, I am not sure what the average user expectation here is 🤔 I think when I add a movie to my watchlist I would prefer to be redirect to the watchlist page afterwards. If I want to see the movie details I can click on it in the watchlist, should be at the first position and easy to find. Same case when adding a new movie play, I think usually would be more interested to see my latest history afterwards.

You know, I agree completely. I changed it so when you add a play you get redirected to history, and a watchlist item – to your watchlist.

I toyed with the idea of just refreshing the page if you were already looking at the page you were redirected to (in case you had e.g., filters active) — but I think this would be counterintuitive if you added a movie to your watchlist but did not see it because of active filters

> > have the "/movies/12" page redirect to e.g., "/movies/12-the-brasher-doubloon"
> 
> great idea!

I will look into this in a separate thread ;]

…

with this change (no longer linking to *film* but just to parent history/watchlist page), this is no longer dependent on #678, so could happily be merged now

leepeuker:
> > I think we need to streamline something more here.
> > There are two UI ways to add a movie to a watchlist:
> > 
> > * use the `Add movie` button -> redirects to watchlist afterwards
> > * use the `Add to watchlist` button on the movie detail page in the `More` modal -> no redirect, reloads the movie page
> > 
> > There are two UI ways to add a movie play:
> > 
> > * use the `Add movie` button -> redirects to history afterwards
> > * use the `Add play` button on the movie detail page -> redirects to the history
> > 
> > The behaviors of the `Add movie` buttons are the same for watchlist/plays, but the behaviors on the movie detail page are conflicting. I think when doing an action on a movie on its detail page it is better to stay on the page instead of redirecting to the history/watchlist, as the user navigated to the movie detail page on purpose instead of using the `Add movie`. What do you think?
> 
> you describe a "redirect from modal" and "do not redirect from page" sense of the world. I think it makes lots of sense.
> 
> At the HEAD of the branch `add/redirect-on-add-to-watchlist` (this PR), what you describe is reality.
> 
> So… no changes have to be made to make what you suggest the case! It already is ^.^

No not exactly as far as I see :thinking:  Currently, when you click on the `Add play` button on a movie detail page and then the `Add play` button in the modal, you will be redirect to the history page. That is what I meant, here we should stay on the page, vs when using the modal via the `Add movie` button, in that case we want to be redirected to the history

leepeuker:
> You're right, don't know how that slipped my testing :S — I will change that to be how is described

Awesome, thank you :)

PR #687, opened 2025-09-08
Add "on watchlist" notifier

alifeee:
this is broken out from #678

add a (simple) notifier to show if a film is on a user's watchlist

image

### button not obvious

in my opinion, the presence of an "add" or "remove" button is not the same as being shown the state of something (whether a film is on your watchlist)

image

### button not visible publically

secondly, the "add to watchlist"/"remove from watchlist" button is not visible when looking at a public profile not logged in as that user, so a different way of showing if a film was on the watchlist would be good for public viewers

image

## design

the design is very simple as I wanted to gauge whether this would be a useful feature first. if it is, then I can work on a better design to make it suit the page more.

### DIFF
--- a/templates/page/movie.html.twig
+++ b/templates/page/movie.html.twig
@@ -23,6 +23,12 @@
         <input type="hidden" id="movieId" value="{{ movie.id }}">
         <input type="hidden" id="username" value="{{ routeUsername }}">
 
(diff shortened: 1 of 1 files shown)


leepeuker:
closed because of https://github.com/leepeuker/movary/pull/691

PR #688, opened 2025-09-08
redirect from e.g., `/movies/14` to `/movies/14-emma`

alifeee:
extracted from https://github.com/leepeuker/movary/pull/685#issuecomment-3263287504

#678 allows "name slugs" like TMDB, such that films URLs can look like:

`http://movary.test/users/alifeee/movies/19-starship-troopers`

This PR adds a redirect, so that visiting e.g.,

`http://movary.test/users/alifeee/movies/19`

…will result in a 301 redirect to the former URL

## bugs

I expect adding this might result in some amount of bugs, so it should not be added lightly.

For example, there is already a bug if you visit e.g., `http://movary.test/users/alifeee/movies/19-tony-christie--going-home` you get directed to `http://movary.test/users/alifeee/movies/19-tony-christie--going-home-starship-troopers-starship-troopers-starship-troopers-starship-troopers-starship-troopers-starship-troopers-starship-troopers-starship-troopers-starship-troopers-starship-troopers-starship-troopers-starship-troopers-starship-troopers-starship-troopers-starship-troopers-starship-troopers-starship-troopers-starship-troopers-starship-troopers-starship-troopers` in a loop…

…oops. I'll fix that one and have a good think what else could go wrong.

### DIFF
--- a/src/HttpController/Web/Middleware/MovieSlugRedirector.php
+++ b/src/HttpController/Web/Middleware/MovieSlugRedirector.php
@@ -37,8 +37,7 @@ class MovieSlugRedirector implements MiddlewareInterface
                 . "/movies/"
                 . $movieId
                 . "-"
-                . $movieTitleSlug
-                . "-",
+                . $movieTitleSlug,
             );
         }
 


alifeee:
this is now ready for review

## change

Requesting a movie or person URL without the "proper" slug returns a 301 (permanent) redirect to the page with the "correct" slug.

Anyone depending on the old functionality (e.g., scraping the movies/persons pages) will have to follow redirects (e.g., `curl -L`) or their tools might be broken. Otherwise, this shouldn't impact the normal browsing experience.

Old request:

```bash
$ curl -I http://movary.test/users/alifeee/movies/14
HTTP/1.1 200 OK
Server: nginx/1.18.0 (Ubuntu)
Content-Type: text/html; charset=UTF-8
```

New request:

```bash
$ curl -I http://movary.test/users/alifeee/movies/14
HTTP/1.1 301 Moved Permanently
Server: nginx/1.18.0 (Ubuntu)
Location: /users/alifeee/movies/14-freakier-friday
```

## some "tests"

```bash
# bad slug
$ curl -I http://movary.test/users/alifeee/movies/14-wrong-slug
HTTP/1.1 301 Moved Permanently
Server: nginx/1.18.0 (Ubuntu)
Location: /users/alifeee/movies/14-freakier-friday

# missing slug
$ curl -I http://movary.test/users/alifeee/movies/14-
HTTP/1.1 301 Moved Permanently
Server: nginx/1.18.0 (Ubuntu)
Location: /users/alifeee/movies/14-freakier-friday

# malformed name
$ curl -I http://movary.test/users/alifeee/movies/14-Freakier-ﬀriday
HTTP/1.1 301 Moved Permanently
Server: nginx/1.18.0 (Ubuntu)
Location: /users/alifeee/movies/14-freakier-friday
```

## notes

this feels like a "dangerous" change that might result in a bug or two, especially as it is a core part of the service (the movie page).

as far as I can tell, there are no bugs. would be good if you had a thunk about it @leepeuker 

this is mostly a quality-of-life fix

leepeuker:
Most of your arguments are not against putting the code into the middleware as far as I see but about the implementation of it. Most would be resolved by refactoring and improving the code I think.

I put this code into the middleware because it seemed fitting to me from an architecture perspective. We have a Middleware to handle request preprocessing, like e.g. checking if the requested url slug is correct. Our actual Controller class is only responsible for the actual business logic, like fetching, mapping and returning the movie data. 
I agree that my implementation is not great. There is code duplication and unnecessary processing, but there are ways to fix this, e.g. by only fetching the required data in the middleware (the movie title) instead of the whole object or fetching the whole movie and caching it for the processing later. The two 404 paths is a good example as well, we could change this to just one and handle the movie not found later in the controller.

What do you think about this view?

leepeuker:
>  It also makes more sense to me to deal with a request "in one go" (in one 'long' function), rather than middleware (which makes sense if it can be reused like the auth middleware)

I guess this is probable the core difference of our perspectives. I have a very OOP focused perspective on the code, a long method is a sign usually a sign that something is not separated cleanly by concern/domain :D 

> anyway, I don't have a strong enough opinion ^.^ — thanks for refactoring it into a way which you think makes sense and is future-proof

Same here, I just did what felt natural and responsible to me. If I would have known the amount of discussion that change would cause I would properly have kept it like it was :D

PR #696, opened 2025-09-16
add SVG posters for movies missing a poster

### LINKED ISSUE #676: Add movie names (especially for movies with missing posters)

alifeee:
closes #676 

please read https://github.com/leepeuker/movary/issues/676#issuecomment-3275774192 and https://github.com/leepeuker/movary/issues/676#issuecomment-3275778597

additional things on those comments are:

- added a 30 day caching header so the SVGs aren't loaded on every page-load. Cache isn't forever in case someone adds the image to TMDB

:]

### DIFF
--- a/public/js/app.js
+++ b/public/js/app.js
@@ -205,7 +205,7 @@ function loadLogModalSearchResults(data) {
         }
 
         backdropPath = item.backdrop_path != null ? 'https://image.tmdb.org/t/p/w780' + item.backdrop_path : null;
-        posterPath = item.poster_path != null ? 'https://image.tmdb.org/t/p/w92' + item.poster_path : APPLICATION_URL + '/images/placeholder-image.png';
+        posterPath = item.poster_path != null ? 'https://image.tmdb.org/t/p/w92' + item.poster_path : APPLICATION_URL + '/images/placeholder/' + btoa(item.title)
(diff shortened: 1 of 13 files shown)


leepeuker:
> added a 30 day caching header so the SVGs aren't loaded on every page-load. Cache isn't forever in case someone adds the image to TMDB

When a movie gets a TMDB poster the path of the image changes, so the cache duration should not be relevant as we are loading a new image. 

I think the main issue is changes to the placeholder image svg itself. Worst case: A user sees two different placeholder images for 30 days (user loads placeholder image A, then deploys new release with design changes, then adds new movie with placeholder image B => image A and B would look different for 30 days). We could reduce the caching time to e.g. 1 day as long as we are not on a stable version to plan ahead for this case and increase the caching? What do you think?

leepeuker (approved):

PR #697, opened 2025-09-16
move homepage to `/users/<user>` from `/users/<user>/dashboard`

alifeee:
I find myself trying to visit this 404 page

`http://movary.test/users/alifeee`

it makes sense to me for that to redirect to

`http://movary.test/users/alifeee/dashboard`

…which is what this PR adds

### DIFF
--- a/settings/routes.php
+++ b/settings/routes.php
@@ -178,6 +178,7 @@ function addWebRoutes(RouterService $routerService, FastRoute\RouteCollector $ro
     ##############
     # User media #
     ##############
+    $routes->add('GET', '/users/{username:[a-zA-Z0-9]+}[/]', [Web\DashboardController::class, 'redirectToDashboard'], [Web\Middleware\IsAuthorizedToReadUserData::class]);
     $routes->add('GET', '/users/{username:[a-zA-Z0-9]+}/dashboard', [Web\DashboardController::class, 'render'], [Web\Middleware\IsAuthorizedToReadUserData::class]);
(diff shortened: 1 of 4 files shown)


leepeuker:
> perhaps it is worth thinking whether the /dashboard/ section of the URL is needed ? why should a user dashboard not just be served on the root…

That is a good question. Let's change that and serve the dashboard at the root of the user path. Maybe let us add a redirect for the old dashboard url for backwards compatibility

alifeee:
> That is a good question. Let's change that and serve the dashboard at the root of the user path like you suggested. Maybe let us add a redirect for the old dashboard url for backwards compatibility

should be an "easy enough" change. might miss some places where the URL is built with `dashboard`, but a 301 redirect should solve that.

another choice here is that we could add this line to the `nginx` configuration which would redirect trailing slashes

```nginx
rewrite ^(.+)/+$ $1 permanent;
```

…so, currently, `http://movary.test/users/alifeee/` is a 404, but it could instead 301 redirect to `http://movary.test/users/alifeee`. this depends on what people expect, though, I suppose. I think it's nice as trailing slashes are quite common so typing the URL in with a trailing slash to get a 404 is a little unexpected

leepeuker:
Let us split the dashboard url change and the trailing `/` into separate branches/discussions, as these are separate topics.

alifeee:
hm I could have sworn I fixed e024e84b353c511b32cbafcbc77b034e00c57f8f, but anyway thanks for ;]

PR #710, opened 2025-10-14
Feature/add movary link to tooltips

alifeee:
this PR replaces this…

> I have also added a "movary" link. You see it on the images above. It is a hyperlink which takes you to the movie page from the search, i.e., /users/alifeee/movies/124

…from https://github.com/leepeuker/movary/pull/709

it must wait for https://github.com/leepeuker/movary/pull/709 until merging

now the watchlist/played icons themselves link to the movie page, see:

image

this doesn't add any extra visual elements, and is probably a bit of a secret. But I want to do it, so I made it :laughing:


alifeee:
> Can add link to the icons on the search result list as well? I have added tooltips there https://github.com/leepeuker/movary/pull/711, too.

I feel that would be confusing, especially on mobile, as the whole rectangle is clickable, so would be weird to have internal links that do something unexpected

image

PR #706, opened 2025-10-09
FEAT: cross-posting to Mastodon

alifeee:
add `Mastodon` settings page and other logic in prep for cross-posting to mastodon

Done:

- modify user database table with parameters
- create settings page
- create test button on settings page
- add "post to mastodon" checkbox to log modal
- pass "post to mastodon" bool to movie play/watchlist logging PHP

to-do:

- create job to post to mastodon
- trigger job
- test

## context

Mastodon allows API access to accounts on https://mastodon.social/settings/applications/

image

we can post statuses to Mastodon with:

```bash
TOKEN="8uqah2tr78hwauhfguihweauihguihesuigh973eh7u8ighe"
curl "https://mastodon.social/api/v1/statuses" \
  -H "Authorization: Bearer ${TOKEN}" \
  -H "Idempotency-Key: ${RANDOM}" \
  -F "status=test post using the API: ${RANDOM}" \
  -F "visibility=private"
```

## this PR

this PR adds this settings page `/settings/integrations/mastodon`, where you can enable cross-posting, input an API access token, input a username (which is used for the domain), choose what post type should be used, and enable automatic posting (see below)

<details><summary>the names of these settings are inconsistent, expand:</summary>

(due to so many files needing to be edited and the structure/nature of the application)

```text
in SQL
mastodon_xpost_enabled
mastodon_xpost_automatic
mastodon_xpost_username
mastodon_xpost_access_token
mastodon_xpost_post_visibility

in PHP
$mastodonXPostEnabled,
$mastodonXPostAutomatic,
$mastodonXPostUsername,
$mastodonXPostAccessToken,
$mastodonXPostPostVisibility,

in Twig        
mastodonEnable
mastodonOnByDefault
mastodonUsername
mastodonVisibility
mastodonAccessToken
```

</details>

image

then, on the log play modal, we get a checkbox (only visible if cross posting enabled, default value is from automatic)

image

### on jobs

currently, the settings are implemented, but the logic is not.

`WatchlistController->addMovieToWatchlist` and `HistoryController->logMovie` should create a job to post a Mastodon status. I've not done this yet. Jobs are confusing.

it also seems to me that: `movary-job-processing` runs every 15 seconds and processes the oldest job. this is no good if the rate of job creation ever goes above 1 every 15 seconds

### on watchlist adding

I have no plans to currently trigger Mastodon from the movie page as "add to watchlist" button uses completely different logic to the one on the log play modal (sad)

### on jobs (again)

[...]

PR #722, opened 2025-11-06
add og-data to templates

alifeee:
which shows richly on social media sites when shared — see https://ogp.me/

there are many `og-data` checkers online to check, but in brief it consists of these tags

```html
    <meta property="og:title" content="title goes here" />
    <meta property="og:type" content="website" />
    <meta property="og:url" content="https://url.goes.here" />
    <meta property="og:image" content="https://image/goes/here" />
    <meta property="og:description" content="description goes here" />
    <meta property="og:locale" content="en_GB" />
```

I've added all but the last one (as we do not know the language of a site…? or maybe it's always "English" because there is no internationalisation)

before:

image

(some sites use the `<title>` HTML tag to display a title)

after

image

## notes

- primarily I have done this so that in https://github.com/leepeuker/movary/pull/706 when the URL is in a Mastodon post there is a movie poster automatically shown
- I have used the movie poster URL directly from the page template
- The `description` is not changed per-page, currently. I couldn't really think of a good reason why the current one isn't just fine.

### DIFF
--- a/templates/base.html.twig
+++ b/templates/base.html.twig
@@ -8,7 +8,7 @@
     <title>{% apply trim %}{% block title %}Movary{% endblock %}{% endapply %} | {{ applicationName }}</title>
     <link rel="shortcut icon" href="{{ applicationUrl }}/images/favicon.ico" type="image/x-icon">
 
-    <meta property="og:title" content="{{ applicationName  ~ ': ' ~  block('title')|trim }}" />
+    <meta property="og:title" content="{{ block('title')|trim }} | {{ applicationName }}" />
     <meta property="og:type" content="website" />
     <meta property="og:url" content="{{ applicationUrl }}{{ canonicalPath }}{{ canonicalExtra }}" />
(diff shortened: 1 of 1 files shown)


alifeee:
> With my pushed adjustments I would be fine to merge, please have a look and if you agree with the changes feel free to merge

thanks for adjusting – it's nicer having only one `title` block ! I have adjusted your adjustments, let me know what you think

Firstly I moved the "keyword" for the page to the start. This makes tab titles more obvious…

before

Screenshot from 2025-11-07 13-33-08

after

Screenshot from 2025-11-07 13-42-55

…and also means that the ownership `'s` does not have to fit every username (as it would not)

----------

I also added ` | Movary` (or other name) to the end of tab titles

image

let me know what you think… happy to merge :]

leepeuker:
> Firstly I moved the "keyword" for the page to the start. This makes tab titles more obvious…

I see your point, but I have reverted this again, sorry :D It reads/sounds weird to with "Dashboard for USER", maybe it should be "Dashboard of USER", which seems still not really fitting to me as well. I have compared what other sites are doing (mainly letterboxed and trakt.tv) and they are using the username first as well, so I guess this is not untypical :D

PR #744, opened 2026-03-28
Fix broken pages of movies with numeric names

leepeuker:


alifeee:
presumably this also breaks for movies that don't contain any characters from `[a-z0-9]` (e.g., non-Latin script characters)

perhaps it's better to allow empty "nice" URLs than to allow only `[a-z0-9]`

PR #774, opened 2026-08-23
Add sahinakkaya into contributors

sahinakkaya:


sahinakkaya:
can I ask why? @leepeuker

leepeuker:
Hey @sahinakkaya, I only wanted to explicitly list contributors in the README which made multiple contributions to the project or over a longer amount of time. GitHub itself already lists all contributors automatically in the sidebar (which includes you), I did not want to duplicate this feature in the README and instead focus on "bigger" contributors there. As your PR did not contain any more information (like a PR comment why you want to be added now) I simply closed it without a comment as well. To be honest, I thought this was a LLM PR without much thought behind it, I get quiet a few of those the last few months :sweat_smile: 

If you still think you should be added to the README contributors I am open to hear your argument/s :)

sahinakkaya:
Well, I recently realized there is a contributors section in README and my name is not listed there. I checked https://github.com/leepeuker/movary/graphs/contributors?from=01%2F01%2F2021&to=23%2F08%2F2026 and saw that pbogre is listed in contributors section in the README with "less" contributions. I am not arguing my contributions worth more. I just checked raw numbers and saw that my addition/deletion count is higher so I thought it would be fair if I also want to put my name in the README. 

It was not an automated pr from LLM so my heart was kinda broken when you closed it without a comment :D But since we've both explained ourselves now, I think there are no hard feelings. I understand your reasoning and it is up to you if you want to merge it or not.

leepeuker:
> I just checked raw numbers and saw that my addition/deletion count is higher so I thought it would be fair if I also want to put my name in the README.

True, but for example the number of commits is lower, the question is how do we want to measure it :D 

As we have no "official" rules here and you took your time to explain yourself I am fine adding you to the list. Thank you for your understanding (and contributions :P).

